PR #13, opened 2015-08-17
Push local fixes

jfryman:
Before attempting to test out some changes with the build pipeline, I saw there were some local changes to the repository. So, grabbed those and put them in a commit.


manasdk:
+1

PR #15, opened 2015-08-18
Reduce noise on ChatOps workflowsl

jfryman:
This PR reduces noise for a few of the workflows that are used in ChatOps. Namely:
- `st2cd.create_vm_role`
- `st2cd.destroy_vm`.

I'm not a fan of how tightly coupled the action meta now is to the workflow itself, but this is what we have.

### DIFF
--- a/actions/create_vm_role.meta.yaml
+++ b/actions/create_vm_role.meta.yaml
@@ -58,3 +58,16 @@
       description: "used by rule with actiontrigger"
       immutable: true
       default: "slack"
+  skip_notify:
+    default:
+      - get_subnet_id
+      - get_ami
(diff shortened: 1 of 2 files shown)


actions/create_vm_role.meta.yaml
```diff
@@ -58,3 +58,16 @@
       description: "used by rule with actiontrigger"
       immutable: true
       default: "slack"
+  skip_notify:
```
lakshmi-kannan, line 61: This looks good. We can think about getting this from aliases meta as well.

lakshmi-kannan:
l,g,t,m

PR #18, opened 2015-08-19
increase timeout of puppet runs to 10 minutes

jfryman:
This PR increases the runtime of a puppet converge from 60 seconds to 10 minutes.


jfryman:
Fixes runtime errors. https://stackstorm.slack.com/archives/opstown/p1439943869000892

PR #24, opened 2015-08-28
Release pipeline fixes

manasdk:
1) Fix st2client release rule - actiontrigger and specifically on that action provide repo in parameters. The prvious would've worked with a git postcommit hook most likely.

2) st2cd.st2_pkg_ubuntu14 does not have distro parameter - Likely a relic from deploy_test or something where distro is supplied


m4dcoder:
+1. Thanks for fixing this.

DoriftoShoes:
+1

PR #25, opened 2015-08-28
RFR: We already cded to the cover_dir, so just use style.css

lakshmi-kannan:


DoriftoShoes:
+1 So how many man hours did it take for us to suss this one out?

lakshmi-kannan:
So many. Haha.

PR #26, opened 2015-08-28
Let's not much around with the mime type.

DoriftoShoes:


actions/s3_docs.sh
```diff
@@ -11,6 +11,6 @@ cd $REPO
 
 if [ "$#" -gt 0 ]; then
     for LOCATION in $@; do
-        s3cmd put --recursive * s3://${BUCKET}${LOCATION}
+        s3cmd put --no-mime-magic --guess-mime-type --recursive * s3://${BUCKET}${LOCATION}
```
Kami, line 14: So much magic :P

PR #42, opened 2015-10-09
Pytests breakdown

lakshmi-kannan:


Kami:
We always run pytests on the same server, right?

If so, we should use pip caching capabilities (if we don't yet). I believe those should also be enabled by default in recent version and should work out of the box unless home directory is nuked after each build.

actions/workflows/pytests_split.yaml
```diff
@@ -0,0 +1,44 @@
...
+    -
+      name: "make_compile"
+      ref: "core.remote"
+      params:
+        cmd: "cd {{repo_target}} && make compile"
+        hosts: "{{build_server}}"
+        timeout: 60
+      on-success: "make_reqmnts"
+    -
+      name: "make_reqmnts"
```
Kami, line 12: Common, it's not that hard to spell out the whole word :P

actions/workflows/pytests_split.yaml
```diff
@@ -0,0 +1,44 @@
...
+      params:
+        cmd: "cd {{repo_target}} && make .lint"  # .flake8 and .pylint
+        hosts: "{{build_server}}"
+        timeout: 180
+      on-success: "make_tests"
+    -
+      name: "make_tests"
+      ref: "core.remote"
+      params:
+        cmd: "cd {{repo_target}} && make .pytests"  # unit and integration tests
```
manasdk: .pytests-coverage ?

Atleast as per https://github.com/StackStorm/st2/blob/master/Makefile#L238

lakshmi-kannan:
> We always run pytests on the same server, right?

We load balance across a bunch of build slaves but still we should cache. I'll look into this next.

actions/workflows/st2_pytests.yaml
```diff
@@ -40,11 +40,10 @@
...
     -
       name: "pytests"
-      ref: "st2cd.make_pytests"
+      ref: "st2cd.pytests_split"
       params:
-        hosts: "{{build_server}}"
-        repo: "{{repodir}}"
-        timeout: 650
+        build_server: "{{build_server}}"
+        repo_target: "{{repo_target}}"
```
manasdk, line 46: repodir?

lakshmi-kannan, line 46: No. repodir is just st2_master_`date +%s`. We want the full path.

PR #46, opened 2015-10-13
Start st2workroom_test on every commit to st2workroom/master

manasdk:
- Only run ubunutu14 for now
- hostname includes truncated commit sha which can lead to hostname
  collisions when teh push event is replayed from github.

Also once we are all happy I will enable the webhook on st2workroom (https://github.com/StackStorm/st2workroom/settings/hooks/6096740) and then magic shall happen.


jfryman:
woot!

manasdk:
I take that as a +1. Merging!

PR #50, opened 2015-10-15
A way to push packages to npm and push corresponding tags to github

manasdk:


actions/npm_publish_version.meta.yaml
```diff
@@ -0,0 +1,25 @@
...
+  description: "Published new versions to npm."
+  enabled: true
+  entry_point: "workflows/npm_publish_version.yaml"
+  parameters:
+    repo:
+      type: "string"
+      description: "Url of the repo to clone"
+      default: "git@github.com:StackStorm/hubot-stackstorm.git"
+      required: true
+    branch:
```
Kami, line 13: This will only operate on a release version branch like elsewhere, right?

(Just want to confirm, since operating on non-release branch wouldn't work well without also specifying a commit sha)

actions/workflows/npm_publish_version.yaml
```diff
@@ -0,0 +1,43 @@
...
+        target: "{{repo_target}}"
+      publish:
+        repodir: "{{clone_repo[build_server].stdout}}"
+      on-success: "npm_publish"
+    -
+      name: "npm_publish"
+      ref: "core.remote"
+      params:
+        hosts: "{{build_server}}"
+        cmd: "cd {{repodir}} && npm version {{update_type}} && npm publish"
```
Kami, line 29: I assume we want to increment the version ourselves since `npm version` is naive and doesn't know which version you want to increment (major / minor / patch).

jfryman, line 29: What is the safety if I try and tag a module with something I already have? Also, is there an opportunity to leverage Git Tags here? (for example, pass in a git tag and that is what gets published, and we use that as version?)

actions/npm_publish_version.meta.yaml
jfryman, line 13: See question re: tags

actions/workflows/npm_publish_version.yaml
Kami, line 29: Yeah, ideally we would pass in an exact revision / tag - this way we know what we are publishing and there is no chance an invalid thing is accidentally (race, push to the wrong branch, etc.) publish.

actions/npm_publish_version.meta.yaml
manasdk, line 13: We ain't got no release branch here and it does not really matter from the npm perspective. 

Basically, I am trying to only capture the process I follow manually and that works on master. If we change the process then this workflow can be modified to match the process. For now though we ship from master!

actions/workflows/npm_publish_version.yaml
manasdk, line 29: > I assume we want to increment the version ourselves since npm version is naive and doesn't know which version you want to increment (major / minor / patch).

Actually the `update_type` is exactly that i.e. we get to pick major/minor/patch and `npm version` assuming semantic versioning does the right thing. The has worked well for me manually so I simply replicated that approach.

> What is the safety if I try and tag a module with something I already have?

Don't tag manually - that my current process :)

> Also, is there an opportunity to leverage Git Tags here?

`npm version` actually creates a tag matching the new version locally. So we are leveraging git tags except that `npm version` is performing the tagging so its not obvious. Btw, notice the `git_commit_updates` task in the workflow - thats where the tag gets pushed.

manasdk:
@Kami & @jfryman  - I do not think I really understand the context of your questions. I suspect this is because you guys have a mental model of the release process in this case and I may not have explicitly shared mine. So to spark the what is the right process conversation let me first describe mine -
1. Find a suitable build server that has all dependencies(npm) satisfied. Refer https://github.com/StackStorm/st2puppet/pull/190
2. Clone the branch with a node module like `hubot-stackstorm`.
3. Bump up version locally following semantic versioning schemes. Therefore we pick major, minor or patch as one of the update types and let `npm version` do its magic.
4. Publish to the npm registry with `npm publish`.
5. Commit changes to `package.json` back to git repo
6. Push locally created tag to git repo

This is literally the process I followed to manually push new releases and is pretty much the verbatim description of the process. What are the changes you guys would like to see - do bear in mind that the tooling might come with some of its own quirks and behaviors and for better or worse I would rather not fight with the tooling :)

manasdk:
Hmm, maybe I should simply close this one. No one seems to care much about the PR.

jfryman:
@manasdk maybe you could just push what you have, and then we learn and iterate.

Also, passive aggressive, much? Come on, friend. :heart:

manasdk:
Not going to use this so closing.

PR #53, opened 2015-10-20
Change the bootstrap steps to use curl like the instructions

manasdk:
- Using http instead of https to deal with cipher mismatches.


manasdk:
I've tested this on st2build002 already so not waiting for a +1

PR #56, opened 2015-10-22
Tests for workroom with st2enterprise answers file

Kami:
This PR adds some basic integration / end to end tests for st2enterprise workroom using answers file.

We test the following things:
1. Verify RBAC is enabled
2. Verify Flow is installed and available at /flow
3. Verify LDAP auth backend is installed and configured

Right now we don't run end to end tests since for that we will need an actual configured LDAP server so the tests can authenticate. That's something we should eventually also do (one thing we could do is use a static and preconfigured LDAP server which we always test against).

Eventually I also plan to move inline shell checks to small bash scripts which will get scped to the server and executed there.


rules/st2workroom_st2enterprise_test_master_el6.yaml
```diff
@@ -0,0 +1,31 @@
...
+        trigger.parameters.environment:
+            pattern: "staging"
+            type: "equals"
+        trigger.parameters.branch:
+            pattern: "master"
+            type: "equals"
+    action:
+        ref: "st2cd.st2workroom_st2enterprise_test"
+        parameters:
+            hostname: "st2w-st2enterprise-{{trigger.parameters.branch}}-u14-{{trigger.execution_id}}"
```
jfryman: Based on the file, it should be `EL6`, but it`s`U14` here.

Kami: lol, I just copied the file over and made the modifications.

Anyway, I will fix the file name.

PR #75, opened 2015-11-03
Allow branches for st2tests

lakshmi-kannan:


manasdk:
So since the parameter passed is optional this will work with older version of the st2-setup-tests script.

Still the old script will pull from master which is an issue.

PR #83, opened 2015-11-11
Use curl | sh to install as that hopefully fails

manasdk:


actions/workflows/st2workroom_st2enterprise_test.yaml
```diff
@@ -56,9 +56,24 @@
           DISABLE_HUBOT: "true"
           CI: "true"
         hosts: "{{hostname}}"
-        cmd: "curl -sSL https://stackstorm.com/install.sh -o /tmp/bootstrap.sh && chmod +x /tmp/bootstrap.sh && /tmp/bootstrap.sh -a /tmp/answers.yaml"
```
Kami, line 59: So previously it didn't fail?

I'm a bit confused since we used && and not ; so it should have failed...

manasdk, line 59: I do not understand either but for whatever reason it did not fail. This does so I stopped fighting :)

PR #93, opened 2015-12-10
Release checklist

emedvedev:


emedvedev:
@dzimine: please review. That pretty much sums up my experience with 1.2.0 sans upstream and build box issues. The checklist will be updated in the future (I'd like to transfer patch release steps here, too, right after I actually release one), but right now it's a good starting point.

PR #96, opened 2015-12-18
Fix trigger type for librato event

manasdk:
Well obviously actiontrigger is the wrong trigger type

### DIFF
--- a/rules/st2_add_librato_annotation_for_release.yaml
+++ b/rules/st2_add_librato_annotation_for_release.yaml
@@ -3,7 +3,7 @@
     description: "Creates librato annotation when a release is created."
     enabled: true
     trigger:
-        type: "core.st2.generic.actiontrigger"
+        type: "GitHubWebhook.github_event"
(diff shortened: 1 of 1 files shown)


manasdk:
@DoriftoShoes @Kami 

Btw, what is the action this rule triggers even supposed to be doing? Should this annotation be added at the end of a release so after finalize or something? Likely would need a bit more work.

We could add at start of release also like the rule is setup to do but that means it would miss out on patches.

manasdk:
On closer inspection I am not sure what this is doing really. The way I have changed it every commit to a release branch will lead to an annotation. Previous rule definition was just wrong.

manasdk:
Will go ahead and merge for now.

Kami:
@manasdk Yep - it should send an annotation to librato on release.

This way we can cross reference other metrics to release, etc.

It definitely worked in the (distant) part, but yeah this have changes around and this rule was probably never updated.

PR #98, opened 2016-01-06
Modify st2_docs workflow

dzimine:
With this, st2_docs can run 'standalone', independent of full st2 build, to make and push docs from any branch to s3.

I removed the stuff and steps that seem to be not used.

Tested by running manually on st2build002. Didn't test as part of the whole build yet.

### DIFF
--- a/actions/st2_docs.meta.yaml
+++ b/actions/st2_docs.meta.yaml
@@ -1,10 +1,11 @@
 ---
   name: "docs"
+  pack: "st2cd"
   runner_type: "action-chain"
   description: "Build st2docs"
   enabled: true
   entry_point: "workflows/st2_docs.yaml"
-  parameters: 
+    parameters: 
     repo: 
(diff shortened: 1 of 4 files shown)


estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1452047607020548

DoriftoShoes:
Looks good.  I'm reasonably sure this will work fine.  The parameters you removed were used for context further in build process but nothing triggers off of the docs jobs anymore.  Just make sure to update the docs rule also to remove those parameters and the build should be fine...w00t.

manasdk:
We should abandon this in favor of https://github.com/StackStorm/st2cd/pull/100

dzimine:
abandoning.

dzimine:
I want to reanimate this and apply to st2docs. So that we can build and push docs manually off any branch.

PR #99, opened 2016-01-07
Fix hostname conflict

dzimine:
Problem: when rules with different distors fire the workflow at the same time, VM creation fails for all but first because the DNS name is already taken.


estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1452157874020880

Kami:
Good catch :+1:

Kami:
I'll go ahead and merge it.

I think it should be fine and the record name is not too long.

PR #101, opened 2016-01-08
WIP: Fix hostname, take 2

dzimine:
Problem - with PR #99 a hostname is too long. 
Per talking to @DoriftoShoes, conforming to existing pattern: form hostname in a rule and pass down to action chain.

PS. I drafted a version with one rule + mistral workflow running 3 versions in parallel but ran out of time to complete; will finish when have more time to play.

### DIFF
--- a/rules/puppet_st2_ci_el7.yaml
+++ b/rules/puppet_st2_ci_el7.yaml
@@ -16,4 +16,5 @@
         ref: "st2cd.workroom_puppet_st2_ci"
         parameters:
             revision: "{{trigger.body.head_commit.id}}"
+            hostname: "st2wp-el7-{{trigger.body.head_commit.id | truncate(8, False, '')}}"
             distro: "RHEL7"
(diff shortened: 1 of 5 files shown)


Kami:
Lol, looks like my gut feeling from here (https://github.com/StackStorm/st2cd/pull/99#issuecomment-169659511) was wrong.

rules/puppet_st2_ci_el7.yaml
```diff
@@ -16,4 +16,5 @@
         ref: "st2cd.workroom_puppet_st2_ci"
         parameters:
             revision: "{{trigger.body.head_commit.id}}"
+            hostname: "st2wp-el7-{{trigger.body.head_commit.id | truncate(8, False, '')}}"
```
dzimine, line 19: @DoriftoShoes @lakshmi-kannan 
I follow @lakshmi-kannan path here using commit ID. Trouble is, re-running it on the same github hook... FORCES me to clean-up AWS zombies :) Keep? Change to using execution-id like @DoriftoShoes was doing?

estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1452269163021639

PR #102, opened 2016-01-08
Add some basic checks which verify that a no default private and public key is installed for stanley

Kami:
This pull request adds some basic checks which verify that a no default private key and public key is installed for stanley user after install.sh script completes.

As an additional check / improvement we should also add some tests which verify that the key is indeed installed when `st2::stanley::ssh_private_key` value is set in the answers file.

### DIFF
--- a/actions/workflows/st2workroom_test.yaml
+++ b/actions/workflows/st2workroom_test.yaml
@@ -76,27 +76,13 @@
         distro: "{{distro}}"
         user: "{{st2_username}}"
         pass: "{{st2_password}}"
-      on-success: "verify_no_default_private_ssh_key_is_installed_for_stanley_1"
+      on-success: "verify_no_default_keys_are_installed_for_stanley_user"
     -
-      name: "verify_no_default_private_ssh_key_is_installed_for_stanley_1"
-      ref: "core.remote_sudo"
+      name: "verify_no_default_keys_are_installed_for_stanley_user"
+      ref: "st2cd.st2workroom_verify_default_stanley_ssh_keys_are_not_installed"
       params:
         hosts: "{{hostname}}"
-        cmd: 'test -f /home/stanley/.ssh/st2_stanley_key && (echo "Private SSH key file exists for stanley user" ; exit 1)'
-      on-success: "verify_no_default_private_ssh_key_is_installed_for_stanley_2"
-    -
-      name: "verify_no_default_private_ssh_key_is_installed_for_stanley_2"
-      ref: "core.remote_sudo"
-      params:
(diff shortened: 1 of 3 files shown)


estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1452276280021714

actions/workflows/st2workroom_test.yaml
```diff
@@ -76,7 +76,29 @@
         distro: "{{distro}}"
         user: "{{st2_username}}"
         pass: "{{st2_password}}"
+      on-success: "verify_no_default_private_ssh_key_is_installed_for_stanley_1"
```
manasdk: Could these test be a separate workflow simply called from here? I suspect these will grow.

actions/workflows/st2workroom_test.yaml
```diff
@@ -76,7 +76,29 @@
...
+      params:
+        hosts: "{{hostname}}"
+        cmd: 'files=$(find /home/stanley/.ssh/ -type f ! -name authorized_keys | wc -l) ; if [ $files -gt 0 ]; then echo "Found SSH key files"; exit 1; fi'
+      on-success: "verify_no_default_authorized_key_is_installed_for_instaley"
+    -
+      name: "verify_no_default_authorized_key_is_installed_for_instaley"
+      ref: "core.remote_sudo"
+      params:
+        hosts: "{{hostname}}"
+        cmd: 'lines=$(cat /home/stanley/.ssh/authorized_keys | grep st2_stanley_key | wc -l) ; if [ $lines -gt 0 ]; then echo "Found default stanley in authorized_keys"; exit 1; fi'
```
manasdk: lol ... this constitutes an abuse in my view :P

Kami: I agree, I'm not a fan of it, but right now there is really no nicer way to do it.

I was thinking about putting those checks in script files and rsyncing / scping those files over and directly calling them.

Kami: Actually, I will try porting it remote runner script action and see how it goes.

I wanted to do that in the past with st2enterprise tests, but I had some issues.

Kami:
I'll go ahead, merge this into master and work from there.

The whole develop / test loop is already slow and having this change in a branch makes it even slower.

actions/workflows/st2workroom_test.yaml
Kami: Much better now - https://github.com/StackStorm/st2cd/pull/102/files#diff-b42acea6d62fe3c15c64402a82b63f13

PR #104, opened 2016-01-12
Re-enable default SSH key checks

Kami:
This should be merged after https://github.com/StackStorm/st2workroom/pull/305 is merged into master otherwise workroom tests will fail.

(Previously I enabled it so I could more easily test those changes on build server)

TODO:
- [x] Make sure we explicitly install a new default SSH key after this check so it can be used by the tests


Kami:
Alright, I merged this into master and the tests finally passed.

selection_319

PR #106, opened 2016-01-12
Include missing /

manasdk:
- Missing `/` mean we end up with `/tmpbootstrap` instead of `/tmp/bootstrap`
- Whitespace was just editor doing its thing

### DIFF
--- a/actions/fetch_puppet_bootstrap.yaml
+++ b/actions/fetch_puppet_bootstrap.yaml
@@ -4,23 +4,23 @@
   description: "Download puppet bootstrap script"
   enabled: true
   entry_point: ""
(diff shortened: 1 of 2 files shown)


Kami:
I'm really confused - how did this ever work?

I checked git history in this and st2puppet repo, but I didn't find any "offending" changes. Or did I miss something?

manasdk:
Just to be clear this change is good but point less.

`/tmpbootstrap`  is a legit path. Anyway will still merge for the sake of sanity :)

Kami:
Yeah, this makes it a bit more obvious. Previous path was confusing :)

PR #107, opened 2016-01-12
Update install URLs to install.stackstorm.com

lakshmi-kannan:
See https://stackstorm.atlassian.net/browse/STORM-1946 for the reasons. We did install.stackstorm.com so we can support cipherlists supported by RHEL6. So let's use the latest and greatest links.

### DIFF
--- a/actions/create_workroom_test_rule.py
+++ b/actions/create_workroom_test_rule.py
@@ -5,11 +5,7 @@ import json
 import requests
 import sys
 
-INSTALL_URLS = {
-    'UBUNTU14': 'https://stackstorm.com/install.sh',
-    'RHEL6': 'https://raw.githubusercontent.com/StackStorm/st2workroom/master/script/bootstrap-st2',
-    'RHEL7': 'https://raw.githubusercontent.com/StackStorm/st2workroom/master/script/bootstrap-st2'
-}
+INSTALL_URL = 'https://install.stackstorm.com/'
 
 
 def create_rule(url, rule):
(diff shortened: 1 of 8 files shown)


actions/create_workroom_test_rule.py
```diff
@@ -51,7 +47,7 @@ def _get_st2_rules_url(base_url):
 
 def _create_distro_rule_meta(distro, branch, action_ref, version_task_index, distro_release=None):
     # last 3 chars is distro short
-    distro_short = distros[len(distros)-3:]
+    distro_short = distro[len(distro)-3:]
```
lakshmi-kannan, line 50: This should have never worked. distros is not the right variable.

DoriftoShoes, line 50: Will this work even now?  (rhel6 & rhel7) or (centos6 & centos7) will get the same shortname?

DoriftoShoes:
All the url changes look fine.  The shortname for a distro thing still looks weird though.  Also, it's kind of hilarious that we were using the stackstorm.com url in some places and the github raw one in others.

actions/create_workroom_test_rule.py
manasdk, line 50: We only create 3 rules. Can create more for Centos6 and Centos7 if needed.

manasdk:
:+1:

lakshmi-kannan:
> Also, it's kind of hilarious that we were using the stackstorm.com url in some places and the github raw one in others.

This was intentionally done when RHEL6 wouldn't work with the SSL cipher offered by stackstorm.com/install.sh

PR #110, opened 2016-01-21
Add enterprise workroom test case which verifies that the default RBAC admin role assignments are present

Kami:
The title says it all.


actions/st2workroom_st2enterprise_verify_default_rbac_admin_role_assignments_are_present.yaml
```diff
@@ -0,0 +1,8 @@
+
+---
+  name: "st2workroom_st2enterprise_verify_default_rbac_admin_role_assignments_are_present"
```
manasdk, line 3: lol this name wins the award for the longest action name ever :)

PR #117, opened 2016-02-01
 End to end test for st2workroom and installer

Kami:
This pull request adds end to end test which also invoke the installer.

They are based on the existing st2workroom st2enteprise test. The only difference is that instead of providing some of the values in answer.yaml file, those values are provided via the installer and installer is also used to trigger the puppet run.


estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1454348505003815

Kami:
Note - I only added "st2enterprise" tests since they are basically a super-set of regular (non enterprise) tests.

estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1454435130004141

Kami:
Added some more checks which verify that the correct public and private key is also installed after the installer run.

rules/st2workroom_st2installer_st2enterprise_test_master_ubuntu14.yaml
```diff
@@ -0,0 +1,31 @@
+---
```
manasdk, line 1: similar rule needs to be auto-generated for versioned release pipeline. It would have to be added in the prepare_release workflow.

Can be a separate PR.

Kami, line 1: For now I'm also fine with skipping those tests in the release pipeline.

Just testing on master should be a good start.

actions/workflows/st2workroom_st2installer_st2enterprise_test.yaml
```diff
@@ -0,0 +1,134 @@
...
+      on-success: "verify_no_default_keys_are_installed_for_stanley_user"
+    -
+      name: "verify_no_default_keys_are_installed_for_stanley_user"
+      ref: "st2cd.st2workroom_verify_default_stanley_ssh_keys_are_not_installed"
+      params:
+        hosts: "{{hostname}}"
+        sudo: true
+      on-success: "run_installer"
+    -
+      name: "run_installer"
```
manasdk, line 80: If we added all these steps to https://github.com/StackStorm/st2cd/blob/master/actions/workflows/st2workroom_st2enterprise_test.yaml after https://github.com/StackStorm/st2cd/blob/master/actions/workflows/st2workroom_st2enterprise_test.yaml#L149 how would it be different?

estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1454436865004288

actions/workflows/st2workroom_st2installer_st2enterprise_test.yaml
```diff
@@ -0,0 +1,145 @@
...
+        cmd: "curl -sSL https://install.stackstorm.com/  | sudo sh -s -- -a /tmp/answers.yaml"
+        timeout: 3600
+      on-success: "verify_no_default_keys_are_installed_for_stanley_user"
+    -
+      name: "verify_no_default_keys_are_installed_for_stanley_user"
+      ref: "st2cd.st2workroom_verify_default_stanley_ssh_keys_are_not_installed"
+      params:
+        hosts: "{{hostname}}"
+        sudo: true
+      on-success: "run_installer"
```
manasdk, line 78: Once `answer.yaml` is present in /opt/puppet/hieradata/ and installer is run AFAICT it writes `answers.json`. Per my understanding `answers.yaml` > `answers.json` in the precedence order and only 1 of those files is read or do we perform a smart merge? 

It is not obvious to me how https://github.com/StackStorm/st2workroom/blob/c6fc74ca90ca063e8fad6756fe2892575a016fd0/hiera.yaml is interpreted or works.

actions/workflows/st2workroom_st2installer_st2enterprise_test.yaml
Kami, line 80: As I mentioned on Slack - we want to test a clean state and not test a VM which is already provisioned.

And we want to test two scenarios from a clean state:
1. Non GUI installer where info is provided in answers.yaml (existing enterprise tests)
2. GUI based installer where part of the info is provided via the installer and second part of the provision is done by the installer (those new tests)

So in short - it would be very different because at that point the VM is already fully provisioned and I don't want to test an already fully provisioned VM - that's not a goal of those new tests.

actions/workflows/st2workroom_st2installer_st2enterprise_test.yaml
manasdk, line 78: Answering my own question; this is automagically done by hiera 

https://docs.puppetlabs.com/hiera/1/configuring.html#hierarchy

rules/st2workroom_st2installer_st2enterprise_test_master_ubuntu14.yaml
manasdk, line 1: Ok. We should try to have them ready for v1.4

actions/workflows/st2workroom_st2installer_st2enterprise_test.yaml
manasdk, line 80: > Non GUI installer where info is provided in answers.yaml (existing enterprise tests)

Is this test unnecessary? I do not see this as a real customer scenario and these tests are customer scenarios per my understanding.

manasdk:
In this context I guess running installer on non-enterprise installer is also a missing scenario. I will add those tests in a separate PR following the example from here - we have seen quite a few failures emerge from that as well. Although in that case I feel we can add to the existing case.

actions/workflows/st2workroom_st2installer_st2enterprise_test.yaml
manasdk, line 80: My personal preference is to have st2workroom_enterprise tested on all supported platforms in customer scenarios. Which means I am actually making a case for 3 separate tests. Could be added in separate PRs.

Kami:
@manasdk Correct.

The reason I only done st2enterprise so far is that enterprise version is really a super-set of non-enterprise version so in theory it should also exercise all of the non-enterprise version code paths :)

But yeah, if we can come up with an approach which is easy to maintain (so not too much code is duplicated between the workflow) having additional tests for non-enterprise version would also be good.

actions/workflows/st2workroom_st2installer_st2enterprise_test.yaml
Kami, line 78: So yeah, magic :P

manasdk:
:+1: 

Lets go ahead and stabilize these test. I can refactor into shareable workflows later and look at the whole picture to decide if we need to tweak testing.

PR #118, opened 2016-02-03
Add policies for retrying st2workroom_* test workflows on failure (up to 1 time)

Kami:
This should hopefully reduce the burden and stress of false positives (and related time consuming work of checking things out and re-running stuff) and help mitigate false positives related to networking and other intermediate / temporary issues (timeout when pulling docker images, timeout when talking to github, etc.).

### DIFF
--- a/policies/retry_st2workroom_st2enterprise_test_on_failure.yaml
+++ b/policies/retry_st2workroom_st2enterprise_test_on_failure.yaml
@@ -1,0 +1,11 @@
+---
+name: st2workroom_st2enterprise_test.retry_on_failure
+# Note: We retry this run on failure to try to avoid false positives
+# which are caused by intermediate networking issues and similar.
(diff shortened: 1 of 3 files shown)


estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1454526991004709

manasdk:
Without proper cleanup this may not work due to hostname clashes. If the policy could re-run with a new param that would have likely solved the issue.

Kami:
True, that's a good point.

@m4dcoder You said you are working on the changes which will try to delete the VM and DNS entry if one already exists, right?

Kami:
Another simply short term solution would simply be to make sure name is not unique - append a couple of random characters to the end of the name.

Any objections to that change?

Kami:
Now that #119 has been merged, I believe I can go ahead and merge this, right?

PR #121, opened 2016-02-06
Check if hostname return from linux.dig in st2cd.destroy_vm

m4dcoder:
If there is no result for the hostname, then exit the workflow.


estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1454719521005370

PR #124, opened 2016-02-09
Workroom upgrade tests workflow

Kami:
This pull request introduces tests for workflow upgrades (WIP).
## TODO
- [x] Write post-upgrade checks and hook it up in the workflow
- [x] More checks in the upgrade checks script (suggestions welcome)
- [x] Test it end to end
## TODO (future improvements)
- [ ] Add rules and hook it up to our release pipeline
- [ ] Retry on failure policy to mitigate false negatives due to intermediate network issues
- [ ] Refactor the code and split it into multiple smaller actions / workflows which can be-reused in other workflows (future improvement)
- [ ] Based it on top of the enterprise tests since those should be a super-set of non-enterprise tests (future improvement)
- [ ] Also run those tests on  RHEL / EL (future improvements)


estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1455017511005878

estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1455059996006170

manasdk:
> More checs in the upgrade checks script (suggestions welcome)

Thinking of adding a test to check if all services are running. e2e should eventually catch this issue anyway so its not a blind spot. Making it explicit is likely to save us time.

Kami:
> Thinking of adding a test to check if all services are running. e2e should eventually catch this issue anyway so its not a blind spot. Making it explicit is likely to save us time.

Yeah, that would be nice.

That's also one of the reasons why I added `st2ctl restart` in the workflow - to make sure all the services start back up after the upgrade.

Kami:
Added checks for services / processes running to the post upgrade checks script.

Kami:
I've managed to get the workflow to pass (execution id 56bb244b82fb9b41f8620917) with some workaroung for st2ctl "bug" / issue.

As mentioned on Slack, it looks like there is a bug in `st2ctl` and action runners are not always killed (we only send SIGTERM, but it looks like we should also implement sending SIGKILL if SIGTERM doesn't kill processes after a while). I believe multiple users have been affected by this issue in the past.

In any case, that's something which should be fixed in the `st2` repo.

manasdk:
Validated for `v1.2.0` -> `v1.3.1` as well as `v1.3.0` -> `v1.3.1` upgrades.

manasdk:
>  Add rules and hook it up to our release pipeline
>  Retry on failure policy to mitigate false negatives due to intermediate network issues

I will add both of these later as part of separate PRs. For v1.3.x will run these manually once production builds are ready. I am not sure what versions we need to validate upgrades from so will need some process discussion.

My guess is -
1. previous stable version to latest. (e.g. v1.2.0 -> v1.3.2)
2. previous patch versions on current major release to latest. (v1.3.0 -> v1.3.2, v1.3.1 -> v1.3.2)

manasdk:
merging .. pushed some todos out to future.

PR #119, opened 2016-02-04
Refactor st2cd.destroy_vm as a mistral workflow

m4dcoder:
Refactor st2cd.destroy_vm as a mistral workflow. The workflow will gracefully exit if no instances is found.

PR #126, opened 2016-02-10
[DONT MERGE] Destroy VM and "retry on failure" policy changes

Kami:
This adds back changes from #119 and #118 which were reverted to get things working / passing.

Keep in mind that we still need to update all the workflows which utilize `ec2_run_instances` because the result format has changed.

This should only be merged after v1.3.1 release is out and after the AWS pack stuff and existing workflows have been updated.


Kami:
Merged into #128, will continue in that PR / branch.

PR #128, opened 2016-02-10
Fix result handling of ec2_run_instance

manasdk:
Fix to align with changes to AWS pack https://github.com/StackStorm/st2contrib/pull/393

### DIFF
--- a/actions/destroy_vm.meta.yaml
+++ b/actions/destroy_vm.meta.yaml
@@ -1,36 +1,29 @@
 ---
-  name: "destroy_vm"
-  runner_type: "action-chain"
-  description: "Destroys a VM and removes it from Route53"
-  enabled: true
-  entry_point: "workflows/destroy_vm.yaml"
-  parameters:
-    environment:
-      type: "string"
-      description: "Environment to deploy to"
-      default: "staging"
-    dns_zone:
-      type: "string"
-      description: "Route53 DNS Zone to add host to"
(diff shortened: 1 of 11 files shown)


Kami:
Cool.

It would also be a good idea to merge the branch from #126 (destroy_vm_changes) into yours or vice-versa so everything is self contained and easy to test.

Kami:
I will merge my PR into this one.

manasdk:
@Kami Your call on how to merge this branch etc.

estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1455171804006873

Kami:
I'm testing those changes and it doesn't look like `destroy_vm` works correctly - it doesn't seem to delete an instance if one already exists.

In any case, I will let @m4dcoder test and investigate this when he's back since he has more context on the whole Mistral workflow...

Kami:
Also, we need to be careful when testing those changes since it would be really bad if the action deleted wrong instance(s).

Kami:
OK, so the good news.

I tested those + aws pack changes locally on the build server and it looks like they work - all the workroom tests pass. There might still be some other tests / workflows affected, but we should fix those when they pop up.

I will go ahead, merge both of those changes and deploy them to st2build002.

"destroy_vm" still doesn't look like it works as it should, but I'll leave this to @m4dcoder when he's back.

PR #135, opened 2016-03-09
Workflow for package promotion from staging to prod

emedvedev:
This is a WIP workflow for promoting packages from staging to prod. It accepts package, version, revision, os, and repo, generates the right download and upload URLs, and uploads the package for one or multiple OSes.

Right now it’s based on `package_cloud`, so it has to be installed on the machine. Also, the token has to go into the workflow.

It’s still WIP because RHEL isn’t working and error reporting could be better. Any other improvements you might have are greatly appreciated.

There’s also a neat chatops alias to go along with it.

### DIFF
--- a/actions/promote_package.yaml
+++ b/actions/promote_package.yaml
@@ -19,13 +19,17 @@ parameters:
     required: true
   os:
     description: OS to promote (will promote all if not specified)
-    required: false
+    required: true
+    default: all
     enum:
     - wheezy
     - jessie
     - trusty
     - el6
     - el7
+    - deb
+    - rpm
+    - all
   repo:
     description: "Repo to upload to: stable/unstable, community/enterprise"
     required: true
--- a/actions/promote_package.yaml
+++ b/actions/promote_package.yaml
@@ -1,0 +1,36 @@
+---
+name: promote_package
+runner_type: mistral-v2
+description: Promote a package to production from staging.
+enabled: true
+entry_point: workflows/promote_package.yaml
+parameters:
+  package:
+    type: string
+    description: Package to promote
+    required: true
+  version:
+    type: string
+    description: Version to promote
+    required: true
+  revision:
+    type: string
+    description: Revision to promote
+    required: true
+  os:
+    description: OS to promote (will promote all if not specified)
+    required: false
+    enum:
+    - wheezy
+    - jessie
+    - trusty
+    - el6
+    - el7
+  repo:
+    description: "Repo to upload to: stable/unstable, community/enterprise"
+    required: true
+    enum:
+    - stable
+    - unstable
+    - enterprise
+    - enterprise-unstable
--- a/actions/workflows/promote_package.yaml
+++ b/actions/workflows/promote_package.yaml
@@ -21,72 +21,53 @@ st2cd.promote_package:
         input:
           cmd: "mkdir -p /tmp/st2-up/{ubuntu/trusty,debian/wheezy,debian/jessie,el/6,el/7}"
         publish:
-          el6_path:    el/6
-          el7_path:    el/7
-          rpm: <% $.package %>_<% $.version %>-<% $.revision %>.x86_64.rpm
-          deb: <% $.package %>_<% $.version %>-<% $.revision %>_amd64.deb
-          source: <% $.repo = 'enterprise' and 'enterprise-staging' or $.repo = 'enterprise-unstable' and 'enterprise-staging-unstable' or $.repo = 'stable' and 'staging-stable' or $.repo = 'unstable' and 'staging-unstable' %>
+          rpm: "<% $.package %>_<% $.version %>-<% $.revision %>.x86_64.rpm"
+          deb: "<% $.package %>_<% $.version %>-<% $.revision %>_amd64.deb"
+          source: "<% $.repo = 'enterprise' and 'enterprise-staging' or $.repo = 'enterprise-unstable' and 'enterprise-staging-unstable' or $.repo = 'stable' and 'staging-stable' or $.repo = 'unstable' and 'staging-unstable' %>"
+          token: "772c93363a047c25a0d113fccc0c0963cfa44a45453ba7e9"
         on-success:
           - download_trusty: "<% $.os in list(None, 'trusty') %>"
           - download_wheezy: "<% $.os in list(None, 'wheezy') %>"
           - download_jessie: "<% $.os in list(None, 'jessie') %>"
-          - download_el6: "<% $.os in list(None, 'el6') %>"
-          - download_el7: "<% $.os in list(None, 'el7') %>"
 
       download_trusty:
         action: core.local
         input:
-          cmd: "wget -p /tmp/st2-up/ubuntu/trusty/ https://packagecloud.io/StackStorm/<% $.source %>/ubuntu/pool/trusty/main/<% $.package[0] %>/<% $.package %>/<% $.deb %>"
+          cmd: "wget -O /tmp/st2-up/ubuntu/trusty/<% $.deb %> https://packagecloud.io/StackStorm/<% $.source %>/ubuntu/pool/trusty/main/<% $.package.substring(0,1) %>/<% $.package %>/<% $.deb %>"
         on-success: upload_trusty
+        on-error: cleanup
       upload_trusty:
         action: core.local
         input:
-          cmd: "package_cloud push StackStorm/<% $.repo %>/ubuntu/trusty /tmp/st2-up/ubuntu/trusty/<% $.deb %>"
+          cmd: "LANG=en_US.utf8 PACKAGECLOUD_TOKEN=<% $.token %> package_cloud push StackStorm/<% $.repo %>/ubuntu/trusty /tmp/st2-up/ubuntu/trusty/<% $.deb %>"
         on-success: cleanup
+        on-error: cleanup
 
       download_wheezy:
         action: core.local
         input:
-          cmd: "wget -p /tmp/st2-up/debian/wheezy/ https://packagecloud.io/StackStorm/<% $.source %>/debian/pool/wheezy/main/<% $.package[0] %>/<% $.package %>/<% $.deb %>"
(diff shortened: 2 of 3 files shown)


emedvedev:
screenshot 2016-03-10 01 37 16

emedvedev:
RHEL is working now. You can also use `deb` or `rpm` in place of OS to promote deb-based or rpm-based packages at once.

dzimine:
The workflow is awesome - it shows all the awesome-ness of the workflow - it worths a blog post. Very simple: here is a problem, here is a solution, here is why we like it (chatops). Just need a "flow" screen shot on this, too.

emedvedev:
The flow screenshot would be really uneventful, but yeah, gotta do it if I ever get around to writing a post on this.

actions/workflows/promote_package.yaml
```diff
@@ -0,0 +1,100 @@
...
+      - os
+      - repo
+    tasks:
+
+      init:
+        action: core.local
+        input:
+          cmd: "mkdir -p /tmp/st2-up/{ubuntu/trusty,debian/wheezy,debian/jessie,el/6,el/7}"
+        publish:
+          rpm: "<% $.package %>-<% $.version %>-<% $.revision %>.x86_64.rpm"
```
arm4b, line 20: Sadly, this is not always the case.

both deb and rpm package file names could be much more complex, including epoch, different architecture and so on.

For example @dennybaa has `st2enterprise_1.3-1_all.deb` deployed in private repos.
@lakshmi-kannan wants to use `el7`/`el6` in naming for `rpm` packages which is valid case as well:
`st2-1.4dev-5.el7.x86_64.rpm`.

arm4b, line 20: @emedvedev Actually for now real case is @dennybaa case. If you convince him not to use `_all.deb` (I tried once :) - everything will work with such hardcoding way.

If not - you need to fetch PackageCloud API to discover full real filename for specific pkg name, revision, os, repo.
See example API request: https://github.com/StackStorm/st2-packages/blob/master/.circle/packagecloud.sh#L211

actions/promote_package.yaml
```diff
@@ -0,0 +1,40 @@
...
+    description: Version to promote
+    required: true
+  revision:
+    type: string
+    description: Revision to promote
+    required: true
+  os:
+    description: OS to promote (will promote all if not specified)
+    required: true
+    default: all
```
arm4b, line 23: Packages for every platform/os has independent revision discovery and can have unsynced revision number.

In fact, we had this already and it's normal case, because for example: 
- changing number of parallel builds in CircleCI to 1 will generate packages for 1 platform, rather then 5. It means revision is incremented only for that platform/os.
- triggering new build via CircleCI API can generate packages for specific platforms/OSes. See ChatOps: https://github.com/StackStorm/st2-packages/issues/147

All these cases will bring revision number unsync between different OSes.
So it could be a problem here in future when promoting all platforms and assuming they have same revision number.

---

^^ Just FYI to consider. Though it's not a big issue for the moment, however it will definitely bite in future.

emedvedev, line 23: That's ok. That's why there's a choice of OS (or rpm/deb) available, and also why the workflow won't fail hard on upload errors: it'll show that one promotion succeeded and the others failed, nothing will break. It would be nice to make it smarter in the future, of course.

arm4b, line 23: ^^ It means if user promotes package with feature `A` with revision `N` for `all` or `deb` or `rpm` he can promote artifacts from different builds (eg. different features or bugs) for different OSes.

In reality, because of assumption from user's side, it can be really hard to find bug in production (packages from different builds).

^^ That looks like error-prone for me, if you plan to leave this process manual via ChatOps (not part of release machinery only).

arm4b, line 23: Ah, see: https://circle-artifacts.com/gh/StackStorm/st2-packages/607/artifacts/0/home/ubuntu/packages/payload.json
It's already unsync.

It's metadata for latest build. This file lists which `versions`, `revisions` for different platforms build just generated.

actions/promote_package.yaml
```diff
@@ -0,0 +1,40 @@
...
+  version:
+    type: string
+    description: Version to promote
+    required: true
+  revision:
+    type: string
+    description: Revision to promote
+    required: true
+  os:
+    description: OS to promote (will promote all if not specified)
```
dzimine, line 21: What was I smoking...

Give examples for version, and revision. Is it v1.3, 1.3, 1.3.6, etc? 

For revision: in app other workflows `revision` == git sha. In case of packages, it's likely the last number of build, which is 1.3.2-17. So the choice of word is confusing, please find a right one.

actions/promote_package.yaml
emedvedev, line 23: That's pretty neat! Maybe we should use build number? Or use `version` and omit the revision, just use the latest successful build. Wdyt?

actions/promote_package.yaml
emedvedev, line 21: For version and revision it's whatever our own convention is, so it's likely that anyone who's gonna use this workflow already knows it. We can try better wording, but given @armab's comments below, I might be able to simplify the workflow and we won't need revision anymore.

actions/workflows/promote_package.yaml
dzimine, line 20: +1

dzimine:
A thought about the the context of calling this workflow in CI:
- all packages for StackStorm Community are built and put to staging. 
- we run the tests: that deploys ALL latest packages to, say, 3 operation system.
- tests run and, say 2 OS are +1 and 1 OS is -1.
  Now I want to promote "all" packages on 2 "+1" OS, and "none" of packages on "-1" os.
  From this view point, promoting packages one-by-one is inconvenient (as a user, I carry the list of packages in my head).

As for versions / revision. In most cases we want "the latest" "revision" (again is it a right word?). These revision numbers are different for all the packages. There is never good way to guess what are the current most recent versions for each (I am repeating what @armab said already). So, the "most recent" is a good default.

Side note (which I don't know what to do with) is transactional consistency - what's happening if someone is installing while we are in the process of promoting the packages (nothing good), and is there a way to deal with it, e.g., via PackageCloud API.

We need to also call PackageCloud folks and ask them to add this workflow to their api, it is "dah", very common use case. Takers?

arm4b:
Revision number is important if we talk about this workflow as part of CI/CD.

1.) CircleCI produces packages with explicitly listed revision numbers: https://circle-artifacts.com/gh/StackStorm/st2-packages/607/artifacts/0/home/ubuntu/packages/payload.json
2.) Staging box consumes these packages and runs bootstrap scripts with pinned version + revision https://github.com/StackStorm/st2-packages/blob/master/scripts/st2bootstrap-deb.sh#L14 obtained from Circle (previous step).
3.) Staging box promotes packages for these specific versions/revisions (subj workflow).

^^ The important thing: versions and revisions should be pinned along entire process.

If not, - it will hit race condition when tests were done for one revision, while promotion happens for different revision (new `latest`).
Eg. now `latest` is one package, in 5 minutes it could be different one.
Now `latest` package can pass tests, in 5 minutes `latest` package can fail the tests.

actions/workflows/promote_package.yaml
```diff
@@ -0,0 +1,100 @@
...
+    input:
+      - package
+      - version
+      - revision
+      - os
+      - repo
+    tasks:
+
+      init:
+        action: core.local
```
manasdk, line 16: Hmm, we may want to run this as `core.remote` on one of the build hosts. Generally, we avoid using `st2build002` or whatever is the "build server" as the execution environment of any commands. Not always possible ofcourse.

actions/workflows/promote_package.yaml
```diff
@@ -0,0 +1,100 @@
...
+        input:
+          cmd: "mkdir -p /tmp/st2-up/{ubuntu/trusty,debian/wheezy,debian/jessie,el/6,el/7}"
```
manasdk, line 18: Would inject some unique reference like revision etc in the directory name. This way different instances of this workflow will not clash against each other.

actions/workflows/promote_package.yaml
```diff
@@ -0,0 +1,100 @@
...
+      download_wheezy:
+        action: core.local
+        input:
+          cmd: "wget -O /tmp/st2-up/debian/wheezy/<% $.deb %> https://packagecloud.io/StackStorm/<% $.source %>/debian/pool/wheezy/main/<% $.package.substring(0,1) %>/<% $.package %>/<% $.deb %>"
+        on-success: upload_wheezy
+#        on-error: cleanup
+      upload_wheezy:
+        action: core.local
+        input:
+          cmd: "LANG=en_US.utf8 PACKAGECLOUD_TOKEN=<% $.token %> package_cloud push StackStorm/<% $.repo %>/debian/wheezy /tmp/st2-up/debian/wheezy/<% $.deb %>"
```
manasdk, line 53: better to use `env` property to pass in environment variables.

actions/workflows/promote_package.yaml
```diff
@@ -0,0 +1,100 @@
...
+      download_trusty:
+        action: core.local
+        input:
+          cmd: "wget -O /tmp/st2-up/ubuntu/trusty/<% $.deb %> https://packagecloud.io/StackStorm/<% $.source %>/ubuntu/pool/trusty/main/<% $.package.substring(0,1) %>/<% $.package %>/<% $.deb %>"
+        on-success: upload_trusty
+#        on-error: cleanup
+      upload_trusty:
+        action: core.local
+        input:
+          cmd: "LANG=en_US.utf8 PACKAGECLOUD_TOKEN=<% $.token %> package_cloud push StackStorm/<% $.repo %>/ubuntu/trusty /tmp/st2-up/ubuntu/trusty/<% $.deb %>"
```
manasdk, line 40: is `package_cloud` guaranteed to be present on the box running this command?

You may have to update the corresponding profile in https://github.com/StackStorm/st2puppet.

DoriftoShoes, line 40: Yeah, we will need to make sure the build server has it.

PR #146, opened 2016-03-17
Feeding packagecloud token to promotion workflow

emedvedev:
As @m4dcoder requested, either specifying the token as a parameter or getting it from the KV store.

### DIFF
--- a/actions/promote_package.yaml
+++ b/actions/promote_package.yaml
@@ -8,6 +8,7 @@ parameters:
   token:
     type: string
     description: Packagecloud API token
+    default: "{{system.pkg_cloud_token}}"
   package:
     type: string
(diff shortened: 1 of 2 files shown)


actions/promote_package.yaml
```diff
@@ -5,6 +5,9 @@ description: Promote a package to production from staging.
 enabled: true
 entry_point: workflows/promote_package.yaml
 parameters:
+  token:
+    type: string
+    description: Packagecloud API token
```
m4dcoder, line 10: Can you set default value to get from the datastore? i.e. {{system.pkg_cloud_token}}

m4dcoder:
LGTM! Thanks!

PR #152, opened 2016-04-08
Extending the VM creation workflow and alias

emedvedev:
1. Added `purpose` and `creator` parameters to the `create_vm_role`: they'll be written as tags to the created instance so that we don't have to guess what do we have `tyrion` or `svetlana001` for. Purpose is optional, creator equals to `api_user` if called from chatops or `user` otherwise.
2. Extended the alias to have a nicer syntax (including purpose) on creation, and a prettier output.
3. Added `ubuntu`, `centos6` and `centos7` as enum options to `distro` because `new pkg_CENTOS6 box ellie-test4 for testing the VM creation alias` is a bit weird.

`!ubuntu vm my_hostname for doing things` would be the shortest version of the alias, and there's some regex magic with optional words to make it more human-like, as always.

screenshot 2016-04-08 11 10 37

screenshot 2016-04-08 11 11 22


actions/workflows/create_vm.yaml
```diff
@@ -2,17 +2,15 @@
   chain:
     -
       name: "get_subnet_id"
-      ref: "st2cd.kvstore"
+      ref: "st2.kv.get"
```
manasdk, line 5: Do we know if stbuid002 has the st2 pack by default? Better not to change this if not absolutely needed.

manasdk:
Overall looks good. I am -1 on "Use `st2.kv` instead of `st2cd.kvstore`" unless it is necessary which is not obvious to me.

actions/workflows/create_vm.yaml
Kami, line 5: I quickly checked https://github.com/StackStorm/st2puppet/tree/production/hieradata/ and it doesn't seem so, so we probably need to update it if we want to use that pack.

manasdk, line 5: In that case please make sure that change is picked up as well and all this works e2e on st2build002. Often its simpler to handle that as a separate task but that is totally upto emedvedev.

actions/workflows/create_vm_role.yaml
```diff
@@ -43,13 +41,13 @@
...
-      on-success: "add_name_tag"
+      on-success: "add_tags"
     -
-      name: "add_name_tag"
+      name: "add_tags"
       ref: "aws.ec2_create_tags"
       params:
         resource_ids: "{{ec2_instance_id}}"
-        tags: "Name={{hostname}}"
+        tags: "Name={{hostname}},Creator={{creator}},Role={{role}}{% if purpose %},Purpose={{purpose[0]|upper}}{{purpose[1:]}}{% endif %}"
```
Kami, line 50: I think `title` filter / function should also work :)

Kami:
LGTM

actions/workflows/create_vm_role.yaml
emedvedev, line 50: Title capitalizes every word :( There's also "capitalize", but it puts everything else in lowercase. So had to go with what you see now :)

emedvedev:
`st2cd.kvstore` didn't actually work on my test box, and doesn't seem to work on a clean-slate st2 installs unless I'm missing something. It kinda feels redundant anyway since we have `st2.kv.*`, and I'd rather install `st2` pack on build boxes if nobody has objections.

manasdk:
@emedvedev https://github.com/StackStorm/st2cd/blob/master/actions/kvstore.yaml

Its likely not worth chasing this one down for yourself locally. Feel free to install `st2` pack on st2build002; there is a good bit of value in doing this anyway. 

If you do choose to go ahead with the `st2cd.kvstore` -> `st2.kv.*` change you are on the hook for the changes to st2puppet, validating everything works on st2build002 i.e. making sure that this change does not break st2build002. You can either get this all to work before you commit this PR to st2cd or you go back to using `st2cd.kvstore` and worry about it some other day. All of this depends on your motivation and the time you have on hand.

emedvedev:
Good input. I'll take a stab at updating the `st2build` role, but if I can't get it working in reasonable time I'll just revert to `st2cd.kvstore`.

emedvedev:
Updated st2puppet and st2build002, made a test run with the new workflow (reference ID: 570b881282fb9b12cf2a16c9), everything's working and the VMs are tagged correctly now. `st2build002` has no ChatOps though, so Ellie is the only bot having a working VM alias right now. We can add this to Estee Tew later, maybe

PR #153, opened 2016-04-09
[DO NOT MERGE] Username and password for unattended bootstrap

emedvedev:
Comes with https://github.com/StackStorm/st2-packages/pull/262


emedvedev:
Merging this PR first, since the current bootstrap script will just pass on unknown arguments, so it's safe.

Kami:
LGTM as long as it doesn't explode on unknown args as you have said.

PR #158, opened 2016-04-13
Dynamically assign unique VM name

m4dcoder:
Make VM name unique since more than one workflow instances of st2_pkg_e2e_test from the same trigger type can be running.


Kami:
:+1:

PR #159, opened 2016-04-15
Move new packaging pipeline components to st2ci

manasdk:


manasdk:
Corresponding add https://github.com/StackStorm/st2ci/pull/7

PR #160, opened 2016-04-15
remove old promote action

manasdk:


m4dcoder:
i believe there's a chatops alias for this action as well.

emedvedev:
There's an alias `promote_package.yaml`. Just make it work with the new action (should be easy, input is almost the same), and the old one can burn. :fire:

manasdk:
@emedvedev @m4dcoder Unfortunately same input and almost the same are not the same thing :(

It might be best to rewrite this alias. It will take me a while.

manasdk:
I decided to take the easy way out and moved promote_package https://github.com/StackStorm/st2ci/pull/8

@m4dcoder I know this is contrary to your opinion but it is the easier option. Rather rewrite the alias but I am not motivated to do that right now.

PR #161, opened 2016-04-22
Remove old pipeline

manasdk:
Removing all the rules that comprise the old CD pipeline. This way we will keep the WFs and Action but not have any executions.


DoriftoShoes:
YOLOSHIP. LGTM.

manasdk:
I cannot help but think I might have overzealously removed some extra rules. Well we shall find out :)

PR #169, opened 2016-05-31
Add missing docs rules for version branches

Kami:


manasdk:
v1.3 was create by a meta action so not going to merge that since it messes with the processes.

PR #172, opened 2016-06-17
Refactor release process and workflows

m4dcoder:


actions/git_delete_branch.yaml
```diff
@@ -0,0 +1,34 @@
+---
+name: git_delete_branch
```
Kami, line 2: Just curious, I don't see this action being used anywhere yet - where do we plan to use it?

m4dcoder, line 2: For manually deleting branches that were created when I'm testing the workflows.

PR #187, opened 2016-07-13
Allow specific enterprise version to be installed

lakshmi-kannan:
https://stackstorm.atlassian.net/browse/STORM-2211


actions/st2_upgrade_to_enterprise.sh
```diff
@@ -20,22 +21,56 @@ if [ "${RELEASE}" = "unstable" ]; then
...
-sed -i -e '/\[system\]/ p' -e '/\[auth\]/,/\[system\]/ d' /etc/st2/st2.conf
+install_enterprise_bits() {
+    # Install enterprise bits from packagecloud
+    echo "Downloading from repo ${REPO}..."
+
+    if [[ ${DISTRO} = \UBUNTU* ]]; then
+        curl -s https://${LICENSE_KEY}:@packagecloud.io/install/repositories/StackStorm/${REPO}/script.deb.sh | bash
+        if [[ -z $VERSION ]]; then
+            apt-get install -y st2flow
+            apt-get install -y st2-auth-ldap
```
DoriftoShoes, line 49: Is the st2enterprise package still broken or should we switch to that instead of the individual packages?

DoriftoShoes:
Might want to switch to using st2enterprise package but other than that LGTM

actions/st2_upgrade_to_enterprise.sh
lakshmi-kannan, line 49: No one seems to know how to generate it. We don't have a 1.5.1 st2enterprise package in the repo :/

PR #192, opened 2016-07-21
Append hostname to end of 127.0.0.1 localhost in /etc/hosts

lakshmi-kannan:
http://stackoverflow.com/questions/8625029/why-does-pythons-socket-getfqdn-return-a-long-string-that-looks-like-an-ipv6

```
stanley@st2-ent-pkg-stable-u14:~$ cat /etc/hosts
127.0.0.1 localhost

# The following lines are desirable for IPv6 capable hosts
::1 ip6-localhost ip6-loopback
fe00::0 ip6-localnet
ff00::0 ip6-mcastprefix
ff02::1 ip6-allnodes
ff02::2 ip6-allrouters
ff02::3 ip6-allhosts
127.0.0.1 st2-ent-pkg-stable-u14.uswest2.stackstorm.net
stanley@st2-ent-pkg-stable-u14:~$
```

Two entries for 127.0.0.1 in two different lines. So socket.getfqdn() returns the first one.

```
stanley@st2-ent-pkg-stable-u14:~$ cat /etc/hosts
127.0.0.1 localhost st2-ent-pkg-stable-u14.uswest2.stackstorm.net
```

fixes it.

### DIFF
--- a/actions/set_hostname.sh
+++ b/actions/set_hostname.sh
@@ -19,9 +19,11 @@ if [[ ! ${SUPPORTED_DISTROS[@]} =~ (^| )${DISTRO_LCASE}($| ) ]]; then
 fi
 
 if [[ ${DISTRO_LCASE} = "ubuntu" ]]; then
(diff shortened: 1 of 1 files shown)


DoriftoShoes:
+1

PR #190, opened 2016-07-19
create_vm_role.yaml in mistral and checks for CI

humblearner:

PR #193, opened 2016-07-22
Add output to create_vm_role workflow

m4dcoder:


actions/workflows/create_vm_role.yaml
```diff
@@ -28,16 +39,16 @@ st2cd.create_vm_role:
             input:
                 key: ami_<% $.distro %>
             on-success:
-                - check_cname_if_it_already_exists
```
humblearner, line 31: I kept this as `check_cname` but was asked to change to `check_cname_if_it_already_exists` by @lakshmi-kannan.

m4dcoder, line 31: lol.  i would ignore in this case as i'm doing.

m4dcoder, line 31: just to clarify, i feel the task name is over descriptive.  check_cname is pretty descriptive to me already.

m4dcoder, line 31: And if we want more explanation, we can always add comments at the top of the task so there's no need to make the task name unnecessary long...

humblearner:
Also with `task-default` section at the beginning. We should be overriding the `on-error` task-defaults for tasks before `run-instance` right ? As per your comment: https://github.com/StackStorm/st2cd/pull/190#discussion_r71618330. Just trying to get a better picture of how YAQL works.

m4dcoder:
IMHO, it doesn't matter.  st2cd.destroy_vm is idempotent.  This makes the code cleaner.

humblearner:
Indeed thats the case, code is cleaner. But all the tasks will now make extra call to `destroy_vm`on failure, even when it doesn't exist, prolonging the workflow by little. But as you wish to implement. Thanks for clarifying. :)

m4dcoder:
There's only 1 task (wait_for_instance) that's making the extra call and it's only invoking on error in which the percentage is low which is acceptable.

m4dcoder:
>  all the tasks will now make extra call to destroy_vm on failure

@humblearner: I think you're missing some information.

PR #194, opened 2016-07-25
Updating build url

DoriftoShoes:
We have no way of retrieving the hostname programmatically anymore so I am hardcoding it.


m4dcoder:
LGTM

PR #199, opened 2016-08-11
Allow boxes to be spun up in 'public' subnet

lakshmi-kannan:
Sometimes we want the ability to spin a node in public subnet to share the box with someone else or just expose a demo node. This PR allows one to specify a "public" environment. We detect the subnet based off that key. The subnet id is set correctly in st2-build.


DoriftoShoes:
+1

PR #210, opened 2016-10-04
Disable test_key_triggers

lakshmi-kannan:
Stable packages and unstable packages both use the same tests. Unfortunately, we changed the scope from system to st2kv.system and therefore we can't get this test to pass on both stable and unstable unless we introduce branches in st2tests. So for now, let me disable this test and deal with branching next.


m4dcoder:
@lakshmi-kannan I think you can leave the test in there.  Just use on-complete instead of on-success after test_quickstart_key_triggers is completed.

PR #214, opened 2016-10-29
Fix e2e tests

emedvedev:
Check out packs from git without using packs.install. We should have at
least one place in e2e where we try out that approach.


humblearner:
👍

PR #219, opened 2016-11-09
Add a rule for v2.0 st2docs

lakshmi-kannan:
```
lakshmi@st2-build-node-1:~$ st2 rule-enforcement get 5823995827263a06d242e9c7
+---------------------+-----------------------------+
| Property            | Value                       |
+---------------------+-----------------------------+
| id                  | 5823995827263a06d242e9c7    |
| rule.ref            | st2cd.st2_docs_v20          |
| trigger_instance_id | 5823995627263a06d242e9c0    |
| execution_id        | 5823995727263a06d242e9c6    |
| failure_reason      |                             |
| enforced_at         | 2016-11-09T21:47:03.450532Z |
+---------------------+-----------------------------+
lakshmi@st2-build-node-1:~$ st2 execution get 5823995727263a06d242e9c6
id: 5823995727263a06d242e9c6
action.ref: st2cd.st2_docs
parameters:
  author: LindsayHill
  branch: v2.0
  docs_url: docs.stackstorm.com
  environment: production
  repo: https://github.com/StackStorm/st2docs.git
  revision: bf35c74ccc124c90c61d35b835adcece9d26719d
status: running (29s elapsed)
start_timestamp: 2016-11-09T21:47:03.512051Z
end_timestamp: None
+--------------------------+-------------------------+------------------+-----------------+-------------------------------+
| id                       | status                  | task             | action          | start_timestamp               |
+--------------------------+-------------------------+------------------+-----------------+-------------------------------+
| 5823995927263a069211355c | succeeded (4s elapsed)  | get_build_server | linux.dig       | Wed, 09 Nov 2016 21:47:05 UTC |
| 5823995e27263a069211355e | succeeded (11s elapsed) | clone_repo       | st2cd.git_clone | Wed, 09 Nov 2016 21:47:10 UTC |
| 5823996927263a0692113560 | succeeded (5s elapsed)  | version          | core.remote     | Wed, 09 Nov 2016 21:47:21 UTC |
| 5823996e27263a0692113562 | succeeded (5s elapsed)  | update_url       | core.remote     | Wed, 09 Nov 2016 21:47:26 UTC |
| 5823997327263a0692113564 | running (2s elapsed)    | make_docs        | st2cd.make_docs | Wed, 09 Nov 2016 21:47:31 UTC |
+--------------------------+-------------------------+------------------+-----------------+-------------------------------+
lakshmi@st2-build-node-1:~$
```

PR #222, opened 2016-11-22
docs not minor versioned for bwc

LindsayHill:
Same change as #219, but for BWC docs.

Problem is that if we have BWC v2.0.1, and we update docs, it does not trigger docs rebuild/push to S3. Docs only have x.y versioning, not x.y.z.

Thoughts @lakshmi-kannan ?

### DIFF
--- a/actions/workflows/bwc_docs.yaml
+++ b/actions/workflows/bwc_docs.yaml
@@ -26,7 +26,7 @@
       ref: "core.remote"
       params:
         hosts: "{{build_server}}"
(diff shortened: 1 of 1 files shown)


LindsayHill:
Again, we have discrepancies between bwc-docs.brocade.com/changelog.html and docs.stackstorm.com/changelog.html. I think it's the same problem, due to not handling the minor version. Can we get this change merged?

lakshmi-kannan (approved):

PR #185, opened 2016-07-12
Install bc as test_timer_rules depends on it

lakshmi-kannan:
Hand installed bc on a RHEL7 box for testing and it's running e2e right now:
http://st2build002.uswest2.stackstorm.net:8080/#/history/5785762d82fb9b2df9f0e254/general

PR #231, opened 2016-12-04
Remove `bc` package dependency

arm4b:
See: https://github.com/StackStorm/st2tests/pull/87 which replaces `bc` -> `awk` that already used in the same workflow for calculation.

Also :stuck_out_tongue:  https://github.com/StackStorm/st2cd/pull/185


humblearner:
👍

PR #234, opened 2016-12-06
cd to LOCAL_REPO instead of PROJECT

lakshmi-kannan:
https://st2-build/#/history/58463c5927263a6028fab726/code


lakshmi-kannan:
Fail.

PR #237, opened 2016-12-06
Changing rules to new version.

bigmstone:


Kami:
I think some of those rules work of master and shouldn't have version (the ones which didn't have it before the change), but I could be wrong.

Besides that, LGTM

bigmstone:
Per slack convo we assume those missing needed to get added.

PR #238, opened 2016-12-07
Fix dev version for bwc related pkg suites

m4dcoder:
An additional "dev" was appended to the end of the dev version string.


Kami:
:+1:

PR #239, opened 2016-12-09
Adding action to automate rule creation for release.

bigmstone:
This PR adds automation for rule creation for ST2CD and ST2CI repos.

- [x] ST2CI Rule Creation
- [x] ST2CD Rule Creation
- [x] Parent Workflow to run CI and CD rule creation


bigmstone:
@m4dcoder In other rules you push directly to master on `st2`, etc. repos. It's probably easier to just do that here as well. What do you think? My original thought was to tie into GitHub API and create PR for branch, but that might be more work than is worth it.

actions/workflows/st2_prep_release_rules.yaml
```diff
@@ -0,0 +1,71 @@
...
+    description: Prepare st2cd and st2ci rules for next release.
+    input:
+        - version
+        - prev_version
+        - host
+        - cwd
+
+    task-defaults:
+        on-error:
+            - cleanup_on_failure: <% $.local_repo_sfx != null %>
```
m4dcoder, line 15: Can you add a vars section and init local_repo_sfx to null?  Otherwise, if there's failure with init task before the publish, this task-defaults will fail.

```
    input:
        ...

    vars:
        local_repo_sfx: null

    task-defaults:
        ...
```

m4dcoder:
LGTM other than a minor comment.  Which other rules are you referring to?  Sorry, I don't recall ATM.

bigmstone:
@m4dcoder The other rules I had in mind are the ones that promote versions. So bringing `master` to `2.2dev` for example.

PR #245, opened 2017-02-22
Fixing st2docs prep release

### LINKED ISSUE #244: prep_release_for_st2docs version.txt update error
In `prep_release_for_st2docs` action the grep on line 74 will always match. The compare on line 75 will thus always pass and not update the version correctly and the action will fail.

bigmstone:
Closes #244


bigmstone:
@m4dcoder Does this now operate as you intended? The way it previously was written it would never actually update the version.

Mierdin:
👍  LGTM

PR #246, opened 2017-02-22
Remove another marshmallowkittens and implement robot mistral tests

Mierdin:
This just continues the great work started in https://github.com/StackStorm/st2cd/pull/243, and also enables robot testing for mistral


Kami:
Thanks, looks l missed some spots when fixing  this one - https://github.com/StackStorm/st2cd/pull/243

PR #248, opened 2017-02-22
Using www for google URLs

Mierdin:
https://github.com/StackStorm/st2cd/pull/246 rewrote URLs to use google.com, but unfortunately this results in 301s, which the action treats as a failure.

Using www.google.com instead produces a successful code


bigmstone (approved):
:+1:

PR #249, opened 2017-02-23
Modified gitref sed statements to properly replace commented out vars

Mierdin:
https://github.com/StackStorm/st2-packages/pull/428 adds commented-out definitions for ST2 and mistral gitrefs, so that at release time, these can be replaced with the right gitrefs.

This PR fixes the sed regular expression within the release prep scripts to take this into account.

### DIFF
--- a/actions/st2_prep_release_for_st2_pkg.sh
+++ b/actions/st2_prep_release_for_st2_pkg.sh
@@ -101,8 +101,9 @@ fi
 # Update the st2 version at circle.yml
 CIRCLE_YML_FILE="circle.yml"
 echo "Setting version in ${CIRCLE_YML_FILE} to ${BRANCH}..."
-sed -i -e "s/\(ST2_GITREV:[ ]*\).*/\1${BRANCH}/" ${CIRCLE_YML_FILE}
-sed -i -e "s/\(ST2MISTRAL_GITREV:[ ]*\).*/\1${MISTRAL_VERSION}/" ${CIRCLE_YML_FILE}
+# capture group one is to ensure proper indentation. Capture group two is actual var definition
+sed -i -e "s/\(.*\)# \(ST2_GITREV:[ ]*\).*/\1\2${BRANCH}/" ${CIRCLE_YML_FILE}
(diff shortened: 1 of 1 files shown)


Mierdin:
@armab Not sure why it won't let me assign you but if you could take a look that would be awesome

actions/st2_prep_release_for_st2_pkg.sh
```diff
@@ -101,8 +101,9 @@ fi
 # Update the st2 version at circle.yml
 CIRCLE_YML_FILE="circle.yml"
 echo "Setting version in ${CIRCLE_YML_FILE} to ${BRANCH}..."
-sed -i -e "s/\(ST2_GITREV:[ ]*\).*/\1${BRANCH}/" ${CIRCLE_YML_FILE}
-sed -i -e "s/\(ST2MISTRAL_GITREV:[ ]*\).*/\1${MISTRAL_VERSION}/" ${CIRCLE_YML_FILE}
+# capture group one is to ensure proper indentation. Capture group two is actual var definition
+sed -i -e "s/\(.*\)# \(ST2_GITREV:[ ]*\).*/\1\2${BRANCH}/" ${CIRCLE_YML_FILE}
```
bigmstone: IMO you made this regex too complex (and I love me some regex)

`sed -e "s/# ST2_GITREV: \"\"/ST2_GITREV: ${BRANCH}/" %{CIRCLE_YML_FILE}`

If you wanted to make this more robust (which isn't needed) you should use `\s*` or `\s\+` (Thanks `sed` for all the escapes) to match white space. Matching everything `\(.*\)` is undesirable since it will allow syntax errors when you don't want them without failing.

bigmstone (changes requested):

actions/st2_prep_release_for_st2_pkg.sh
bigmstone: `sed -i -e "s/#\s*\(ST2_GITREV:\s*\).*/\1${BRANCH}/" ${CIRCLE_YML_FILE}`

Mierdin: Okay, updated in https://github.com/StackStorm/st2cd/pull/249/commits/37435cde7c583f6f6d7b6e809775c4f5bf9cc57c

arm4b (approved):
:+1:

PR #250, opened 2017-02-23
Using period instead of 'source' keyword to set up env

Mierdin:
This was causing an error, because this command is running under /bin/sh, which doesn't have the "source" keyword. The dot notation should work across both `sh` and `bash`

> Other tests already use this notation, but we missed this when uncommenting this test in https://github.com/StackStorm/st2cd/pull/246, so this should rectify that


bigmstone (approved):
:+1:

PR #251, opened 2017-02-24
Add update function for mistral upgrade

Mierdin:
Several folks ran into [upgrade issues](https://gist.github.com/enykeev/c58ae6940f88a209e616410c726e002c#gistcomment-1997190) during manual testing and I am running into the same issues when doing automated upgrade testing as part of the release process (same error message):

    error: "(psycopg2.ProgrammingError) column "key" of relation "delayed_calls_v2" does not exist

The existing machinery does seem to perform the migration scripts, but a few `mistral-db-manage` commands need to be run (see [docs](https://docs.stackstorm.com/latest/install/upgrades.html?highlight=upgrades)). So this PR introduces a function to run these commands for each upgrade

### DIFF
--- a/actions/st2_perform_migrations.sh
+++ b/actions/st2_perform_migrations.sh
@@ -145,6 +145,14 @@ run_migration_scripts() {
     fi
   done
 }
+
+update_mistral_db() {
+  sudo service mistral-api stop
+  sudo service mistral stop
+  /opt/stackstorm/mistral/bin/mistral-db-manage --config-file /etc/mistral/mistral.conf upgrade head
+  /opt/stackstorm/mistral/bin/mistral-db-manage --config-file /etc/mistral/mistral.conf populate
+}
+
 ## Version specific migration functions. Note that the function names must have to
 ## migrate_to_${major}.${minor}. Otherwise, those methods won't be run.
 
@@ -153,17 +161,11 @@ migrate_to_1.5() {
   run_migration_scripts $scripts
 }
 
-migrate_to_2.2() {
-  sudo service mistral-api stop
-  sudo service mistral stop
-  /opt/stackstorm/mistral/bin/mistral-db-manage --config-file /etc/mistral/mistral.conf upgrade head
-  /opt/stackstorm/mistral/bin/mistral-db-manage --config-file /etc/mistral/mistral.conf populate
-}
-
 trap 'fail' EXIT
 STEP="Setup args" && setup_args $@
 STEP="Validate versions" && validate_versions
 STEP="Perform migration" && perform_migration
+STEP="Perform updates" && update_mistral_db
 trap - EXIT
 
 ok_message
--- a/actions/st2_perform_migrations.sh
+++ b/actions/st2_perform_migrations.sh
@@ -153,6 +153,23 @@ migrate_to_1.5() {
   run_migration_scripts $scripts
 }
 
+migrate_to_2.2() {
+  sudo service mistral-api stop
+  sudo service mistral stop
+  sudo -u postgres psql
+  \connect mistral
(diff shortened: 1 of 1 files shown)


actions/st2_perform_migrations.sh
```diff
@@ -153,6 +153,23 @@ migrate_to_1.5() {
...
+  DROP TABLE task_executions_v2;
+  DROP TABLE action_executions_v2;
+  DROP TABLE named_locks;
+  \q
+  /opt/stackstorm/mistral/bin/mistral-db-manage --config-file /etc/mistral/mistral.conf upgrade head
+  /opt/stackstorm/mistral/bin/mistral-db-manage --config-file /etc/mistral/mistral.conf populate
+  sudo service mistral start
+  sudo service mistral-api start
+}
+
```
m4dcoder: This is wrong.

m4dcoder: do this instead

```
upgrade_mistral_db() {
    sudo service mistral-api stop
    sudo service mistral stop
    /opt/stackstorm/mistral/bin/mistral-db-manage --config-file /etc/mistral/mistral.conf upgrade head
    /opt/stackstorm/mistral/bin/mistral-db-manage --config-file /etc/mistral/mistral.conf populate
}
```

m4dcoder: also, where is the function being called in this script?

m4dcoder (changes requested):

actions/st2_perform_migrations.sh
Mierdin: Looks like it tries to find a function according to the current upgraded version:

https://github.com/StackStorm/st2cd/pull/251/files#diff-724355fdedcf4c8f85a34f74855378bdL124

That's why I named mine `migrate_to_2.2` - currently, it's not finding an upgrade function so it's exiting with the message `"--> No migrations to run for version $migration_ver"`

m4dcoder: This step is not version specific and has to run for every upgrade.

Mierdin: Okay I pushed 83aa28d to modify the contents of the function - would you still like me to change the name of the function or are we okay with how this is currently executed?

actions/st2_perform_migrations.sh
```diff
@@ -153,6 +153,13 @@ migrate_to_1.5() {
   run_migration_scripts $scripts
 }
 
+migrate_to_2.2() {
```
m4dcoder: Change this to upgrade_mistral_db and add a line at 167 to perform database upgrade.

m4dcoder (changes requested):

actions/st2_perform_migrations.sh
Mierdin: Okay, roger that! So this must be something that is done regardless of version, then

m4dcoder: yes

m4dcoder: I would even run this step before other version specific migration scripts.

Mierdin: Okay, should be gtg in abe127b

m4dcoder (approved):

Mierdin:
@m4dcoder Just for my own curiosity, do we not need to start mistral after this?

```
sudo service mistral start
sudo service mistral-api start
```

m4dcoder:
no, cause there's a restart step after this in the parent workflow

Mierdin:
Doh! You're right, of course :smile: https://github.com/StackStorm/st2ci/blob/master/actions/workflows/st2_pkg_upgrade_e2e_test.yaml#L203

Cool - merging. And thanks!

PR #252, opened 2017-02-24
Changed step to restart specific mistral services

Mierdin:
https://github.com/StackStorm/st2cd/pull/251 implemented a function to perform mistral migrations when doing upgrade testing. However, because of the service dependency, this wouldn't work on all OSs.

This PR rewrites this instead to shut off the two specific mistral services, namely `mistral-server` and `mistral-api`, instead of the umbrella `mistral` service. It also does it on one line because that's how we roll.


bigmstone (approved):
LGTM

PR #260, opened 2017-03-01
Adding action to update bootstrap script with new version

Mierdin:
This PR introduces a new action to replace the branch version being used in the bootstrap script (in st2-packages).

Currently, that line uses single quotes, and so I am just looking for `BRANCH='.*` in my sed regex. Since BRANCH is defined elsewhere, but only here with single quotes, I figured that was useful and perhaps intended. Let me know if this is too fragile and I'll come up with something else.

### DIFF
--- a/actions/st2_update_bootstrap.sh
+++ b/actions/st2_update_bootstrap.sh
@@ -1,0 +1,55 @@
+#!/bin/bash
+set -e
+
+PROJECT="st2-packages"
+BRANCH="master"
+FORK="StackStorm"
+LOCAL_REPO=$1
+GIT_REPO="git@github.com:${FORK}/${PROJECT}.git"
+CWD=`pwd`
+
+
+# CHECK IF BRANCH EXISTS
+BRANCH_EXISTS=`git ls-remote --heads ${GIT_REPO} | grep refs/heads/${BRANCH} || true`
+
+if [[ -z "${BRANCH_EXISTS}" ]]; then
+    >&2 echo "ERROR: Branch ${BRANCH} doesn't exist in ${GIT_REPO}."
+    exit 1
+fi
+
+
+# GIT CLONE AND BRANCH
+if [[ -z ${LOCAL_REPO} ]]; then
+    CURRENT_TIMESTAMP=`date +'%s'`
+    RANDOM_NUMBER=`awk -v min=100 -v max=999 'BEGIN{srand(); print int(min+rand()*(max-min+1))}'`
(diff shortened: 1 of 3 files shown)


actions/st2_update_bootstrap.sh
```diff
@@ -0,0 +1,55 @@
...
+echo "Currently at directory `pwd`..."
+
+# CHECKOUT BRANCH
+if [ "${BRANCH}" != "master" ]; then
+    echo "Checking out branch ${BRANCH}..."
+    git checkout -b ${BRANCH} origin/${BRANCH}
+fi
+
+# Replace BRANCH that uses single-quotes with new version
+sed -i "s/BRANCH='.*/BRANCH='${VERSION}'/g" ./scripts/st2_bootstrap.sh
```
arm4b: This will change `BRANCH=` occurencies in several places in file, including

https://github.com/StackStorm/st2-packages/blob/edba41e5389160a84d4974c85533e5498afb2749/scripts/st2_bootstrap.sh#L87-L95

arm4b (changes requested):

actions/st2_update_bootstrap.sh
arm4b: Ah, I see the trick now:
> 'single quotes'

Anyways, if someone changes the quotes in future PR, - our machinery is doomed :smiley: 

Maybe there is some better approach?

Mierdin: Ack, changed in https://github.com/StackStorm/st2cd/pull/260/commits/5ca917af781f260460570d631d67b4fdd946d213

actions/st2_update_bootstrap.sh
```diff
@@ -0,0 +1,55 @@
...
+echo "Currently at directory `pwd`..."
+
+# CHECKOUT BRANCH
+if [ "${BRANCH}" != "master" ]; then
+    echo "Checking out branch ${BRANCH}..."
+    git checkout -b ${BRANCH} origin/${BRANCH}
+fi
+
+# Replace BRANCH that currently references old version
+sed -i "s/BRANCH=.*${PREVIOUS_VERSION}.*/BRANCH='${VERSION}'/g" ./scripts/st2_bootstrap.sh
```
arm4b, line 46: Just to be sure, do we define/pass `${PREVIOUS_VERSION}` env. variable somewhere?

arm4b: Same for `${VERSION}`, where it's populated?
Couldn't find it during a quick look.

PR #262, opened 2017-03-06
Don't run st2-run-pack-tests tests if they are not available

Kami:
Those tests are only available and applicable to StackStorm >= v2.3dev so they shouldn't run if the tests are not available (e.g. when running tests against an older version branch such as v2.1 or v2.2).

Eventually, we should probably refactor the work-code and move this inside a script is located inside in st2tests repo or similar. This way we don't need to do those checks inside the workflows and the whole thing is more robust.


Kami:
All green and working correctly now (on master and v2.1 branch), merging.

Issue #263, opened 2017-03-22
Update e2e test rules as part of post-release tasks


Mierdin:
All of the rules that perform e2e testing daily are still using an old version of stackstorm:
https://github.com/StackStorm/st2cd/blob/master/rules/st2_pkg_test_stable_u14.yaml#L24.
PRs like https://github.com/StackStorm/st2cd/pull/237 updated these manually, but it's not part of the release wiki, and so it wasn't done for the last few releases.

We need to write an action or workflow to update these automatically as part of the post-release tasks. Someone should probably create workflows to go through all of the rules in st2cd (and st2ci if applicable) that have versions listed in args and update them.

Ideally, this would just get added as a step to `st2_prep_release_rules` as that action already does similar things, and it's already run at the end of both full and patch releases.

Mierdin:
As it turns out, this was already implemented by @bigmstone in https://github.com/StackStorm/st2cd/pull/239. However, it looks like the release immediately following the implementation of that PR either didn't work, or perhaps used the wrong "previous" version, since the script requires a correct previous version in order for the sed replace to work. And I know when I ran 2.2.0, I specified 2.1.1, which would not have matched the sed (as these rules still specify 2.1.0).

So, that leaves a few things left to do:

- [x] https://github.com/StackStorm/st2cd/pull/264 needs to be merged, so that the current rules can reflect the current version (remember to `git pull` and `st2ctl reload` once done).
- [x] Since 2.2.1 release is imminent, I'll debug `st2_prep_release_rules` and ensure that the rules get automatically updated from 2.2.0 to 2.2.1
- [x]  We need to add a step in the wiki to "clean" local `st2cd` and `st2ci` repos before release starts by checking for uncommitted changes, and doing a fresh pull and `st2ctl reload --register-all`. This will be necessary until #256 is implemented

Mierdin:
@bigmstone can you comment on the 3rd TODO from above? Will your build node efforts automatically pull and load the latest `st2ci` and `st2cd` repos?

bigmstone:
@Mierdin the build node will always pull in latest `master` for both repos, yes. That said there's no rule set up to do this on a continual basis. That said the fact that the build node gets destroyed every push to st2 master should remove the issue of people making changes on the server. :)

Mierdin:
That's a bit what I was getting at - basically as long as both repos get a fresh clone every time the build node is built, answers my question. It means I don't have to spend time writing workflows to do this

[...]

PR #264, opened 2017-03-23
Manually update rules to catch up to current release

Mierdin:
The current version in the daily e2e testing rules is out of date, since there's currently no automated process to take care of this during release, and no manual update step is listed in the wiki.

I'll be submitting a PR soon to satisfy https://github.com/StackStorm/st2cd/issues/263, and add automated update steps for future releases, but in order for that to work, the versions as they currently exist needs to be updated to the release that's out as of now, which this PR accomplishes.

**Remember to do a fresh `git pull` and `st2ctl reload` once this is merged, and ensure the new rules are operating**

EDIT: There actually is an automated process, as @bigmstone implemented in https://github.com/StackStorm/st2cd/pull/239. However, it looks like the release immediately following the implementation of that PR either didn't work, or perhaps used the wrong "previous" version, since the script requires a correct previous version in order for the sed replace to work. So, #263 might not be necessary. I'll update there.

### DIFF
--- a/rules/st2_pkg_test_stable_rhel6.yaml
+++ b/rules/st2_pkg_test_stable_rhel6.yaml
@@ -21,4 +21,4 @@ action:
     distro: RHEL6
     pkg_env: production
     release: stable
-    version: "2.1.0"
+    version: "2.2.0"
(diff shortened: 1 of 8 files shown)


lakshmi-kannan (approved):

Mierdin:
Just recording for posterity; I pulled these changes on the build node and reloaded rules:
https://gist.github.com/Mierdin/be6f680cb6a8cf6ad3690651fc70de59

PR #266, opened 2017-03-28
Moving from system to st2kv.system

### LINKED ISSUE #265: Should Move to st2kv

bigmstone:
Fixes #265


humblearner:
Do we need the change here as well ? 

- https://github.com/StackStorm/st2cd/blob/master/actions/workflows/bwc_docs.yaml#L49
- https://github.com/StackStorm/st2cd/blob/master/actions/workflows/st2_docs.yaml#L57

bigmstone:
Good catch @humblearner

humblearner (approved):
👍

Issue #271, opened 2017-04-04
Patch release compatibility for st2_prep_release_rules


Mierdin:
I noticed after the 2.2.1 release, some rules were still working on `2.2.0`:

```
oswaltm@st2-build-node-1:/opt/stackstorm/packs/st2ci$ st2 rule get st2cd.st2_pkg_test_stable_u14
+-------------+----------------------------------------------------------+
| Property    | Value                                                    |
+-------------+----------------------------------------------------------+
| id          | 580455eb27263a5d536f8a2c                                 |
| uid         | rule:st2cd:st2_pkg_test_stable_u14                       |
| ref         | st2cd.st2_pkg_test_stable_u14                            |
| pack        | st2cd                                                    |
| name        | st2_pkg_test_stable_u14                                  |
| description | Test stable packages                                     |
| enabled     | True                                                     |
| action      | {                                                        |
|             |     "ref": "st2ci.st2_pkg_e2e_test",                     |
|             |     "parameters": {                                      |
|             |         "release": "stable",                             |
|             |         "pkg_env": "production",                         |
|             |         "hostname": "st2-pkg-stable-u14",                |
|             |         "version": "2.2.0",                              |
|             |         "distro": "UBUNTU14"                             |
|             |     }                                                    |
|             | }                                                        |
| criteria    |                                                          |
| tags        |                                                          |
| trigger     | {                                                        |
|             |     "type": "core.st2.CronTimer",                        |
|             |     "ref": "core.5b93e4f7-d7d7-40a0-b94f-71712b0c3cc5",  |
|             |     "parameters": {                                      |
|             |         "timezone": "US/Pacific",                        |
|             |         "second": 0,                                     |
|             |         "minute": 0,                                     |
|             |         "hour": 6                                        |
|             |     }                                                    |
|             | }                                                        |
| type        | {                                                        |
|             |     "ref": "standard",                                   |
|             |     "parameters": {}                                     |
|             | }                                                        |
+-------------+----------------------------------------------------------+
```

There is a step in the release wiki that runs the workflow for updating rules to run on the new version:

```
st2 run st2cd.st2_prep_release_rules version=2.1.0 prev_version=2.0.0 -a
```

As I discovered during the release, this works on full releases, but not on patch releases, because the underlying script only compares to minor version, not patch version (i.e. 2.2.0 == 2.2.1)

```
oswaltm@st2-build-node-1:~$ st2 execution get 58e2bad127263a3ed14c1390
[...]

PR #272, opened 2017-04-05
Skip rule creation if already exists

Mierdin:
As explained in https://github.com/StackStorm/st2cd/issues/271, the rule creation step at the end of the patch release was failing because `cat` was throwing an error when the destination file and source file were the same, which will happen for patch releases. This prevented the latter portion of the workflow from running, which updates the st2cd rules to use the new version, which includes patch.

So this PR simply skips the rule creation when these two versions are identical.

Closes https://github.com/StackStorm/st2cd/issues/271


bigmstone (approved):
LGTM :leaves: :egg:

PR #282, opened 2017-06-12
Use cpu-optimized instances where possible

arm4b:
`m3.large` -> `c4.large`

Related to https://github.com/StackStorm/st2ci/pull/89


arm4b:
Seems like old/uneeded workglow

PR #287, opened 2017-06-15
Fix pin dependencies step - make sure python-mistralclient gets added before troveclient

Kami:
troveclient depends on mistralclient so if we don't add mistralclient before troveclient, build will fail (we won't be able to satisfy the dependency needed by troveclient).

### DIFF
--- a/actions/mistral_create_branch.sh
+++ b/actions/mistral_create_branch.sh
@@ -84,7 +84,7 @@ if [[ $(grep -c . <<< "${REQUIREMENTS}") > 1 ]]; then
 
     # Note: Newer versions of troveclient (>=2.10.0) don't work with our mistralclient fork because of the version changing we do
     # See https://github.com/StackStorm/mistral/pull/24 for context and details
-    sed -i "s/^python-troveclient.*/python-troveclient==2.9.0/g" requirements.txts
+    sed -i "s/^python-troveclient.*/python-troveclient==2.9.0/g" requirements.txt
 
     ${GIT} add requirements.txt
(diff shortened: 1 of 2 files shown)


actions/mistral_create_branch.sh
```diff
@@ -81,7 +81,10 @@ if [[ $(grep -c . <<< "${REQUIREMENTS}") > 1 ]]; then
     echo "Updating requirements.txt..."
     REQUIREMENTS=`echo "${REQUIREMENTS}" | sed '/https:\/\/github.com\/stackstorm/Id'`
     echo "${REQUIREMENTS}" > requirements.txt
-    grep -q 'python-mistralclient' requirements.txt || echo "git+https://github.com/StackStorm/python-mistralclient.git@${BRANCH}#egg=python-mistralclient" >> requirements.txt
+    # Note: python-mistralclient must come before troveclient since troveclient depends on mistralclient
+    # In This case we insert it before python-barbicanclient
+    MISTRALCLIENT_LINE="git+https://github.com/StackStorm/python-mistralclient.git@${BRANCH}#egg=python-mistralclient"
+    grep -q 'python-mistralclient' requirements.txt || sed -i "/python-barbicanclient/ i ${MISTRALCLIENT_LINE}" requirements.txt
```
arm4b: what if there is no `python-barbicanclient` string?
then mistralclient is not inserted

Kami: Correct, this could be problematic but we really have nothing else to fall back to.

Not sure it would work correctly if we insert mistral client at the very beginning.

PR #291, opened 2017-06-30
Concurrency++ for 'st2_get_installed_version' WF

arm4b:
Just a random find.

Let the version lookup work in parallel, based on number of packages, instead of running one-after-another with `concurrency: 1`.


Kami:
LGTM :+1: 


(but would be good to test it)

PR #293, opened 2017-07-19
Update tasks in st2_e2e_tests w/f for parallel robot execution

humblearner:


actions/workflows/st2_e2e_tests.yaml
```diff
@@ -69,43 +69,25 @@ workflows:
...
-                    - robot_CLI_tests
-            robot_CLI_tests:
+                    - pabot_chatops_mistral_CLI_tests
+            pabot_chatops_mistral_CLI_tests:
                 action: core.remote
                 input:
                     hosts: <% $.host_ip %>
                     env: <% $.st2_cli_env %>
-                    cmd: cd /tmp/st2tests && . venv/bin/activate && pybot robotfm_tests/cli/
+                    cmd: "cd /tmp/st2tests && . ./venv/bin/activate && bash /tmp/st2tests/robot_parallel_tests.sh '--processes 3 robotfm_tests/cli/ robotfm_tests/mistral/ robotfm_tests/chatopsCI/' 0"
```
arm4b, line 90: From what I understand we put 3 independent tasks into 1 execution:
- robotfm_tests/cli/
- robotfm_tests/mistral/
- robotfm_tests/chatopsCI/

I think they're so different to merge everything in 1 thing.
Can we run them in parallel with Mistral workflows instead and have separated log/task for each execution?

And then it's :+1: to apply pabot for each independent execution, same as you did with `robotfm_tests/docs/`.

I think it will be easier to maintain and navigate in tests for someone from the team.

actions/workflows/st2_e2e_tests.yaml
```diff
@@ -69,43 +69,25 @@ workflows:
...
-                    timeout: 300
-                on-success:
-                    - robot_chatopsCI_tests
-            robot_chatopsCI_tests:
-                action: core.remote
-                input:
-                    hosts: <% $.host_ip %>
-                    env: <% $.st2_cli_env %>
-                    cmd: cd /tmp/st2tests && . venv/bin/activate && pybot robotfm_tests/chatopsCI/
+                    cmd: "cd /tmp/st2tests && . ./venv/bin/activate && bash /tmp/st2tests/robot_parallel_tests.sh '--processes 3 robotfm_tests/docs/' 0"
```
arm4b, line 78: This looks good :+1:, considering the task is logically only one `robotfm_tests/docs/` and we split it.

PR #298, opened 2017-08-02
Update the mistral e2e tasks to run 4(approx) at a time

humblearner:


actions/workflows/st2_e2e_tests.yaml
```diff
@@ -187,8 +178,11 @@ workflows:
                     env: <% $.env %>
                     cmd: st2 run tests.test_packs_pack <% $.st2_cli_args %>
                     timeout: 600
-                # on-success:
-                #     - test_quickstart_local_script
+                on-success:
+                    - test_quickstart_local_script
+                    - test_quickstart_remote_script
+                    - test_quickstart_passive_sensor
+                    - test_quickstart_polling_sensor
```
arm4b, line 185: I get the idea of splitting :+1: 

Did you try to run those 2 failing tests first and once the're finished, - fan-out everything else? That would indicate the race in 2 specific tasks.

Additionally, it would be nice to add a code comment somewhere for every racing task so ideally we could investigate it in future. Race condition in such case probably means something wrong with `st2` itself.

humblearner:
Going to merge this as all the builds are green for last few days!

PR #299, opened 2017-08-11
Remove the robot test for Mistral CLI

m4dcoder:
Remove the test because it is already tested elsewhere.


humblearner (commented):
I do not approve these changes, as the test is already fixed: https://github.com/StackStorm/st2tests/pull/112 and it caught the behavior change when itests for the same didn't.

PR #300, opened 2017-08-24
Increase timeout when uploading docs to s3

m4dcoder:
The command times out at default 60 seconds. Increate the timeout to 600 seconds to give the command more time.


arm4b:
:+1:

PR #301, opened 2017-09-11
Update python path for el6 e2e_st2tests setup

humblearner:
[wheel=0.30.0](https://pypi.python.org/pypi/wheel) removed support for Python2.6.

Fix for e2e_st2tests setup on el6 (Python 2.6.6):
```
Downloading PyYAML-3.12.tar.gz (253kB)
    Complete output from command python setup.py egg_info:
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-build-jqYgYs/PyYAML/setup.py", line 83, in <module>
        from wheel.bdist_wheel import bdist_wheel
      File "/tmp/st2tests/venv/lib/python2.6/site-packages/wheel/bdist_wheel.py", line 407
        ignore=lambda x, y: {'PKG-INFO', 'requires.txt', 'SOURCES.txt',
                                       ^
    SyntaxError: invalid syntax
```


arm4b (approved):
Please include code comment why do we need such hack for EL6.

Besides of that 👍

PR #312, opened 2017-10-25
Fix task name: tag_release -> st2_tag_release

warrenvw:
Change the task name from `tag_release` to `st2_tag_release`, as task `tag_release` doesn't exist.

### DIFF
--- a/actions/tag_release.sh
+++ b/actions/tag_release.sh
@@ -37,7 +37,7 @@ git clone ${GIT_REPO} ${LOCAL_REPO}
 
 cd ${LOCAL_REPO}
 echo "Currently at directory `pwd`..."
-git checkout -b ${BRANCH} origin/${BRANCH}
+git checkout -B ${BRANCH} origin/${BRANCH}
 
 
 # CHECK IF TAG EXISTS
--- a/actions/workflows/st2_finalize_release.yaml
+++ b/actions/workflows/st2_finalize_release.yaml
@@ -117,7 +117,7 @@ st2cd.st2_finalize_release:
         st2_docker_tag_release:
             action: st2cd.tag_release
             input:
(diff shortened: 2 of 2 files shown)


actions/tag_release.sh
```diff
@@ -37,7 +37,7 @@ git clone ${GIT_REPO} ${LOCAL_REPO}
 
 cd ${LOCAL_REPO}
 echo "Currently at directory `pwd`..."
-git checkout -b ${BRANCH} origin/${BRANCH}
+git checkout -B ${BRANCH} origin/${BRANCH}
```
m4dcoder, line 40: Can you explain the difference of -b and -B?  I looked it up with `git checkout -h` but how is it relevant to this patch?

warrenvw: Using the existing `-b`, we see the following error:

```
Cloning into 'st2docker_1508965619_541'...
fatal: A branch named 'master' already exists.
```

This happens because the prior `git clone` points the newly created `HEAD` to the branch pointed to by the cloned repository's `HEAD`.. which by default is `"master"` (effectively creating the local `master` branch). We now need to handle the case where `$BRANCH = 'master'` since this script is also used to tag the `st2-docker` repo. According to the man page: If `-B` is given, `<new_branch>` is created if it doesn't exist; otherwise, it is reset. Reset isn't a problem, since we've just cloned into an empty directory (not overwriting anything in the process).

An even better solution might be using `git clone -b ${BRANCH} ...` and getting rid of the `git checkout` invocation altogether. In this PR, I was trying to go for the least number of changes...

humblearner (approved):
lgtm

arm4b:
Please don't forget to pull the updated repo on live `st2cicd` server.

PR #313, opened 2017-10-26
Update pypi upload API

humblearner:
Thanks @m4dcoder for fix.

### DIFF
--- a/actions/pypi_upload.sh
+++ b/actions/pypi_upload.sh
@@ -59,12 +59,10 @@ index-servers =
     pypitest
 
 [pypi]
-repository: https://upload.pypi.org/legacy/
 username: ${PYPI_USERNAME}
 password: ${PYPI_PASSWORD}
 
 [pypitest]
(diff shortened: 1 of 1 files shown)


arm4b:
Can you clarify in the first message details and background about the issue?

Does it also mean release management servers are able to deploy package to pypi & are fixed or it's a different issue?

m4dcoder:
This change only fix the script to upload to packages to pypi. It does not mean the ubuntu-build servers are fixed. Those servers required the python to be upgrade to latest 2.7.x version for this script to run.

arm4b:
@m4dcoder Thanks for clarification.

If we're not going to fix the `ubuntu-build` servers (python issue) in nearest time, let's please document it as Issue then so next release manager will be aware of it.

Kami (approved):

humblearner:
Updated both `ubuntu-build002` and `ubuntu-build003` from 2.7.6 to 2.7.14 for this script to work

PR #315, opened 2017-11-02
Fix a race between test_quickstart_key_triggers and test_quickstart_python tests

Kami:
There is race condition between `test_quickstart_key_triggers` and `test_quickstart_python` tests so they can't run in parallel.

First test relies on a trigger instance being injected into the system, but second test also injects some trigger instances in the system and depending on the timing, etc. first test can fail because it will retrieve wrong trigger instance.

This pull request changes it to run `test_quickstart_key_triggers` at the end after python one in the sequential manner.

In the future we could ideally refactor the `key_triggers` test so it could still run in parallel, but this would need more rework (right now it just retrieves last trigger instance).

/cc @humblearner

### DIFF
--- a/actions/workflows/st2_e2e_tests.yaml
+++ b/actions/workflows/st2_e2e_tests.yaml
@@ -246,7 +246,6 @@ workflows:
                     timeout: 600
                 on-success:
                     - test_quickstart_python
-                    - test_quickstart_key_triggers
                     - test_quickstart_trace
                     - test_quickstart_run_pack_tests_tool
             test_quickstart_python:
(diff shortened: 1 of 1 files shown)


Kami:
@humblearner If you feel like, it would be great if you can improve `key_triggers` test to avoid the race and tests can still run in parallel.

Here (https://github.com/StackStorm/st2tests/blob/ddb6c8243fcacc03e12cb8739b81b68402705214/packs/tests/actions/chains/test_key_triggers.yaml#L85) we would need to retrieve the actual trigger instance for `key create` and not just the last one.

arm4b:
:+1:

PR #325, opened 2018-04-12
Add a step to kick off Circle CI build to ship API docs

lakshmi-kannan:
This will ship API docs to api.stackstorm.com on every release.

Requires https://github.com/StackStorm/st2apidocs/pull/3 to be merged.


rules/sync_api_docs_on_st2_master_merge.yaml
```diff
@@ -0,0 +1,29 @@
...
+name: sync_api_docs_on_st2_master
+description: Sync API documentation on merges to st2 master
+pack: st2cd
+enabled: true
+
+trigger:
+    type: circle_ci.build_event
+
+criteria:
+    trigger.body.payload.vcs_url:
```
lakshmi-kannan: Won't work. Github changed their payload structure https://developer.github.com/v3/activity/events/types/#pushevent. Getting a branch isn't anymore.

PR #326, opened 2018-04-17
Use pip 9

Kami:
Latest version (10.0.0) might break some things so we should use 9.0.0.1 for now to be on the safe side.

### DIFF
--- a/actions/setup_e2e_tests.sh
+++ b/actions/setup_e2e_tests.sh
@@ -56,13 +56,13 @@ st2ctl reload --register-all
 # Robotframework requirements
 cd st2tests
 sudo pip install --upgrade "pip>=9.0,<9.1"
-sudo pip install --upgrade virtualenv
+sudo pip install --upgrade "virtualenv==15.1.0"
 
 # wheel==0.30.0 doesn't support python 2.6 (default on el6)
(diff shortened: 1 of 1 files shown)


bigmstone (approved):
:+1:

actions/setup_e2e_tests.sh
```diff
@@ -55,7 +55,7 @@ st2ctl reload --register-all
 
 # Robotframework requirements
 cd st2tests
-sudo pip install --upgrade pip
+sudo pip install --upgrade "pip>=9.0,<9.1"
 sudo pip install --upgrade virtualenv
```
arm4b: Need to pin virtualenv as well here (to `15.1.0` which ships with pip 9.0)

arm4b: And same thing, do `virtualenv --no-download` 2 lines below as well, otherwise it will be using latest pip 10.

Kami: Good catch - pushed those changes.

arm4b (approved):
:+1: 

Also a small code comment why we're doing it would be nice, so someone else looking at these lines in future and thinking about upgrading will have a better idea.

Issue #289, opened 2017-06-16
'st2cd' shouldn't set any version variables in K/V


arm4b:
Example: during the release machinery we set `st2_stable_version` in K/V and rely on it.

With the new `st2cicd` server, which could be re-provisioned anytime by us, this becomes a problem to update `ops-infra` Ansible repo with the new K/V value (infra-as-a-code :fist_oncoming: ).
Example: https://github.com/StackStorm/ops-infra/pull/186 :-1: 

The solution is to commit the version used directly to `st2cd` repo itself instead of K/V, same as we do via automation: 
https://github.com/StackStorm/st2cd/commit/2c19af54e6e20c7e7f4a151ef44c9c44c4bae23a

PR #330, opened 2018-04-22
Correctly set stable version for test rules to v2.7.1

Kami:
Some how this didn't get updated during the release process.

For now I will merge that and dig into why those rules were not updated next week.


armab:
Could be https://github.com/StackStorm/st2cd/issues/289.

Because we store version in k/v storage and that doesn't work quite well with `st2cicd` which can be destroyed and reprovisioned anytime.
So instead of saving version in temporary K/V we need to migrate this into infra-as-code, eg. git repo.

PR #332, opened 2018-05-11
Update packer_st2 changelog, commit and tag

warrenvw:
Closes StackStorm/packer-st2/issues/11

Easier and faster to perform the above tasks in one action, although it would be nice to re-use the tag-release action.


actions/workflows/st2_finalize_release.yaml
```diff
@@ -124,8 +124,30 @@ st2cd.st2_finalize_release:
...
+
+        get_utc_date:
+            action: core.local
+            input:
+                cmd: "date -u +%Y%m%d"
+            publish:
+                utc_date: <% task(get_utc_date).result.stdout %>
+            on-success:
+                - update_packer_st2_changelog_and_tag
+        update_packer_st2_changelog_and_tag:
```
arm4b: Let's follow already established naming following the `mistral_tag_release`

eg `vagrant_tag_release` to make it more clear.

actions/workflows/st2_finalize_release.yaml
```diff
@@ -124,8 +124,30 @@ st2cd.st2_finalize_release:
...
+            action: stdcd.update_changelog
+            input:
+                project: packer-st2
+                version: <% $.version + '_' + $.utc_date %>
```
arm4b: Version should be in `v2.7.1-20180131` format

actions/update_changelog.meta.yaml
```diff
@@ -0,0 +1,42 @@
+---
+name: update_changelog
```
arm4b: Because we create git tag (which is more important for release, comparing to changelog), I think we'll need a better naming for this action.

arm4b (changes requested):
Left several comments,

But before merging, please try to run action (from `st2cicd` server) to produce a test release and verify if it works OK.

warrenvw:
I'll test in the morning.

arm4b (approved):
The code LGTM, considering it'll be tested manually.

arm4b:
@warrenvw Great work with the release automation :+1: 
Just a reminder to update the pack on st2cicd in case if you didn't.

PR #340, opened 2018-07-12
Set / bump __version__ for all the runner packages during the release

Kami:
This pull request updates release workflow to set `__version__` for all the runner packages during the release process.

This change is part of https://github.com/StackStorm/st2/pull/4239.

Right now this change is not backward compatible which means we need to decide how to handle backward compatibility for v2.8:

1. Leave this code as is and cherry pick https://github.com/StackStorm/st2/pull/4239 into v2.8 branch (should be fine)
2. Update the workflow so setting runner version is not fatal if runner package `__init__.py` file doesn't contain `__version__` string.

That's fine for v2.8, but I don't like it going forward (it could mask actual issues) - I prefer failing loudly where possible.

### DIFF
--- a/actions/st2_chg_ver_for_st2.sh
+++ b/actions/st2_chg_ver_for_st2.sh
@@ -38,14 +38,18 @@ git clone -b ${BRANCH} --single-branch ${GIT_REPO} ${LOCAL_REPO}
 cd ${LOCAL_REPO}
 echo "Currently at directory `pwd`..."
 
-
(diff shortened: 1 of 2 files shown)


arm4b (approved):
Is think too that, cherry picking (1) sounds good enough & easiest, so we don't overcomplicate it.

Kami:
Alright, I will make that change.

If we ever need to release patch for version < v2.8 we can make corresponding changes to this workflow.

PR #349, opened 2018-10-25
Wait for docker image

warrenvw:
Required since the enterprise image depends on the community image. Without the delay provided by `wait_for_docker_image`, CircleCI will likely attempt to build the enterprise images before the community images are available.


actions/workflows/st2_prep_release.yaml
```diff
@@ -125,6 +125,16 @@ st2cd.st2_prep_release:
                 local_repo: <% 'st2_dockerfiles_' + $.local_repo_sfx %>
                 hosts: <% $.host %>
                 cwd: <% $.cwd %>
+            on-success:
+                - wait_for_community_docker_images
+        wait_for_community_docker_image:
+            action: st2cd.wait_for_docker_image
+            input:
+                org: <% $.fork %>
+                image: st2workflowengine
```
arm4b: Existence of `st2` image would be a more proper check as `st2-enterprise` relies on `stacktorm/st2` base Docker image.

actions/workflows/st2_prep_release.yaml
```diff
@@ -125,6 +125,16 @@ st2cd.st2_prep_release:
...
+                version: <% $.version %>
+                tries: 45
+                default: 60
```
arm4b: What is the `default` parameter and his responsibility for this task?
Can't find it in meta

actions/workflows/st2_prep_release.yaml
```diff
@@ -125,6 +125,16 @@ st2cd.st2_prep_release:
...
                 cwd: <% $.cwd %>
+            on-success:
+                - wait_for_community_docker_images
+        wait_for_community_docker_image:
+            action: st2cd.wait_for_docker_image
+            input:
+                org: <% $.fork %>
+                image: st2workflowengine
+                version: <% $.version %>
+                tries: 45
```
arm4b: Mistral has built-in `retry`: `delay` mechanism which would be nicer to use: https://docs.openstack.org/mistral/ocata/main_features.html#yaml-example
But unsure about Orquesta and so I'm find with current approach unless if it does the job :+1:

arm4b (approved):
Nice one :+1: 

Left one minor comment which looks like oversight in the code.

actions/workflows/st2_prep_release.yaml
arm4b: If it's meant to be `delay`, then waiting for `45tries` x `60secs` which is 45 mins looks like too much.
Let's cap the upper limit to `10mins` or `15mins`. 

According to the `st2-dockerfiles` build, images are usually built in 3 mins, giving 7 more mins should be OK for uploading them to Docker Hub.

actions/workflows/st2_prep_release.yaml
warrenvw: First versions of the code used `st2`, but then I figured since that's not the last image pushed to docker hub we could get into a case where the required image isn't available. However, on further reflection, I'll use `st2web`, since it's the last required image pushed to docker hub.

arm4b: That makes sense :+1:

warrenvw:
```
root@st2vagrant:/opt/stackstorm/packs/st2cd/actions# st2 run st2cd.wait_for_docker_image image=st2web tries=2 delay=5 hosts=localhost version=3.0dev
..
id: 5bd22f92a08f81603d5c53c8
status: succeeded
parameters: 
  delay: '5'
  hosts: localhost
  image: st2web
  tries: '2'
  version: 3.0dev
result: 
  localhost:
    failed: false
    return_code: 0
    stderr: ''
    stdout: SUCCESS!
    succeeded: true
```

warrenvw:
The above vagrant machine is:

```
$ cat /etc/lsb-release 
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=16.04
DISTRIB_CODENAME=xenial
DISTRIB_DESCRIPTION="Ubuntu 16.04.5 LTS"
```

PR #350, opened 2018-11-01
disable VM provision Slack message

LindsayHill:
Reduce some unnecessary monitoring spam

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,0 +1,3 @@
+## 0.1.1 
+
+Disabled rules that report VM provision/destroy messages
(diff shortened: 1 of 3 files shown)


LindsayHill:
Also want to disable "xxx was destroyed" messages and "Unable to deregister monitoring", but will do that once Orquesta conversion is finalized

arm4b (commented):
This debugging log was useful to investigate the CI/CD issues with abandoned/not deleted instances.

Instead of Removing this log completely, would be more elegant to post these messages to `#thunderdome` channel instead of `#opstown`.

That's where all the events from the CI are going and it would be even better context when investigating CI issues.

PR #351, opened 2018-11-02
Opstown spam reduction

LindsayHill:
Also disabled rule for alerting on decommission


arm4b (commented):
Same feedback as https://github.com/StackStorm/st2cd/pull/350#pullrequestreview-171076508

> This debugging log was useful to investigate the CI/CD issues with abandoned/not deleted instances.
Instead of Removing this log completely, would be more elegant to post these messages to `#thunderdome` channel instead of `#opstown`.
That's where all the events from the CI are going and it would be even better context when investigating CI issues.

PR #353, opened 2018-11-14
Reduce opstown noise

LindsayHill:
Further reduction of #opstown noise. 

If/when https://github.com/sangoma/pysensu/pull/11 ever gets merged, we should update the sensu actions to better handle things when deleting a client returns a 404. It's not exactly an error - it means that check does not exist anymore, or was never created.


arm4b:
Also related to https://github.com/StackStorm-Exchange/stackstorm-sensu/issues/3 meaning we can fix it in our Sensu pack.

arm4b (approved):
LGTM

PR #355, opened 2018-11-29
Remove notify_deregister tasks

LindsayHill:
Previously I commented out where we called the "deregister monitoring failed" task. That resulted in an orphan task that always ran. Not my intention. Now commented out the task itself too.


warrenvw:
Looks good to me. Cleaner this way.

LindsayHill:
I've updated this a bit to completely remove the references, rather than just commenting them out.

warrenvw:
👍

PR #358, opened 2018-12-07
Remove pabot tests

blag:
As mentioned on Slack, remove the RobotFM tests that are inconsistently failing.

### DIFF
--- a/actions/workflows/st2_e2e_tests.yaml
+++ b/actions/workflows/st2_e2e_tests.yaml
@@ -95,42 +95,42 @@ workflows:
                     windows_password: <% $.windows_password %>
                     env: <% $.st2_cli_env %>
                     protocol: <% $.protocol %>
-                on-success:
-                    - pabot_docs_tests
-            pabot_docs_tests:
-                action: core.remote
-                input:
-                    hosts: <% $.host_ip %>
-                    env: <% $.st2_cli_env %>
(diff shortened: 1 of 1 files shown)


actions/workflows/st2_e2e_tests.yaml
```diff
@@ -95,42 +95,42 @@ workflows:
...
+            # pabot_chatops_tests:
+            #     action: core.remote
+            #     input:
+            #         hosts: <% $.host_ip %>
+            #         env: <% $.st2_cli_env %>
+            #         cmd: "cd /tmp/st2tests && . ./venv/bin/activate && bash /tmp/st2tests/robot_parallel_tests.sh '--processes 1 robotfm_tests/chatopsCI/' 1"
+            #         timeout: 180
+            #     retry:
+            #         count: 2
+            #         delay: 10
```
m4dcoder: I would remove them altogether.  If we want it back, we can always see earlier commits.

m4dcoder (changes requested):
Minor change request but otherwise LGTM.

actions/workflows/st2_e2e_tests.yaml
m4dcoder: I don't like commenting out codes. This leaves for other people to clean up.

m4dcoder (approved):

actions/workflows/st2_e2e_tests.yaml
blag: Yeah, I wouldn't have left it in if I wasn't planning on ripping them out soonishly.

But either way, fixed.

PR #360, opened 2018-12-13
Make dockerfile workflows idempotent, fix waits

warrenvw:
If the local workspace already existed, then the `st2_chg_ver_for_st2_dockerfiles.sh` script failed. Fixed.

Also added `LOCAL_REPO` where missing.

Please carefully review to ensure the changes are implemented as desired.

### DIFF
--- a/actions/workflows/st2_finalize_release.yaml
+++ b/actions/workflows/st2_finalize_release.yaml
@@ -125,45 +125,7 @@ st2cd.st2_finalize_release:
                 hosts: <% $.host %>
                 cwd: <% $.cwd %>
             on-success:
-                - st2_dockerfiles_tag_release
-        st2_dockerfiles_tag_release:
-            action: st2cd.tag_release
-            input:
-                project: st2-dockerfiles
-                version: <% $.version %>
-                fork: <% $.fork %>
-                local_repo: <% 'st2_dockerfiles_' + $.local_repo_sfx %>
-                branch: master
-                hosts: <% $.host %>
-                cwd: <% $.cwd %>
-            on-success:
-                - wait_for_community_docker_image
-        wait_for_community_docker_image:
-            action: st2cd.wait_for_docker_image
-            input:
-                org: <% $.fork %>
-                image: st2web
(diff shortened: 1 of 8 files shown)


actions/workflows/st2_finalize_release.yaml
```diff
@@ -136,6 +136,19 @@ st2cd.st2_finalize_release:
...
+            on-success:
+                - wait_for_community_docker_image
+        wait_for_community_docker_image:
+            action: st2cd.wait_for_docker_image
+            input:
+                org: <% $.fork %>
+                image: st2web
+                version: <% $.version %>
+                tries: 15
+                check_delay: 60
```
arm4b, line 179: I think `15` mins is too high for this wait.

Looking at https://circleci.com/gh/StackStorm/workflows/st2-dockerfiles average build + push finishes in `4-6 mins`.
I think setting `10mins` would be enough to understand if something went wrong.

actions/workflows/st2_finalize_release.yaml
```diff
@@ -136,6 +136,19 @@ st2cd.st2_finalize_release:
                 branch: master
                 hosts: <% $.host %>
                 cwd: <% $.cwd %>
+            on-success:
+                - wait_for_community_docker_image
+        wait_for_community_docker_image:
```
arm4b, line 172: Additionally, can we move wait for Docker + st2enterprise_tag_release to the end of the workflow, after `ship_api_docs`?

In that case, if something goes wrong with Docker it won't block the other important tasks in the workflow.

actions/workflows/st2_finalize_release.yaml
warrenvw: Sounds good to me.

actions/workflows/st2_finalize_release.yaml
warrenvw: Sure thing.

arm4b (approved):
:+1:

m4dcoder:
LGTM

PR #363, opened 2018-12-18
Additional fixes to the dockerfiles workflows

warrenvw:
A few changes to fix the workflows for `st2-dockerfiles` and `st2enterprise-dockerfiles`.

### DIFF
--- a/actions/st2_chg_ver_for_st2_dockerfiles.sh
+++ b/actions/st2_chg_ver_for_st2_dockerfiles.sh
@@ -5,13 +5,11 @@ set -e
 #  - if missing, create a branch named ${BRANCH} (i.e., "vX.Y") in ${PROJECT}.
 #  - update the Makefile on this branch to reference $ST2_VERSION (i.e., "X.Y.Z").
 
-FORK=$1
-PROJECT=$2
-ST2_VERSION=$3
-LOCAL_REPO=$4
-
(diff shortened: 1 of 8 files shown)


actions/st2_chg_ver_for_st2_dockerfiles.sh
```diff
@@ -8,8 +8,10 @@ set -e
 FORK=$1
 PROJECT=$2
 ST2_VERSION=$3
-BRANCH=$4
-LOCAL_REPO=$5
+LOCAL_REPO=$4
+
+SHORT_VERSION=`echo ${ST2_VERSION} | cut -d "." -f1-2`
+BRANCH="v${SHORT_VERSION}"
```
m4dcoder: I prefer the branch to be passed explicitly into the script.  All of the other st2_cfg_ver_*.sh scripts pass the branch into the script. You can use YAQL expression to get the short version to pass in from the workflow.

m4dcoder (changes requested):

actions/st2_chg_ver_for_st2_dockerfiles.sh
warrenvw: I've addressed your comments in a5a83e6, and also fixed a few additional issues...

m4dcoder (approved):
LGTM. Thanks for being diligent.

PR #365, opened 2019-01-12
Surround version string with quotes

blag:
Prevents ST2 from interpreting the result (`2.10`) as a float, which gets truncated to `2.1`.


m4dcoder:
LGTM. Thank you for fixing this.

m4dcoder (approved):
LGTM. Thank you for fixing this.

PR #368, opened 2019-02-02
Remove newline in version string

warrenvw:
Removing new line so the version is parsed correctly. There should be no space before the trailing forward slash.

```
  version: /2.10 /
```

This PR will resolve the following `s3cmd_docs` error, which has prevented the docs for `2.10` (and other versions) from being installed or updated correctly:

```
    stderr: 'ERROR: Parameter problem: Destination S3 URI must end with ''/'' (ie must refer to a directory on the remote side).'
```

### DIFF
--- a/actions/workflows/bwc_docs.yaml
+++ b/actions/workflows/bwc_docs.yaml
@@ -26,7 +26,7 @@
       ref: "core.remote"
       params:
         hosts: "{{build_server}}"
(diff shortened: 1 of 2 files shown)


Kami (approved):

Kami:
To be on the safe side if we ever run step which writes ``version.txt`` on Windows box (it will likely never happen), we can also remove carriage return character - ``\r``.

PR #379, opened 2019-04-12
Update upgrade to enterprise action to only install st2-rbac-backend package if version is >= v3.0.0

Kami:
This pull request updates upgrade to enterprise action so it only installs new ``st2-rbac-backend`` package if version is ``>= 3.0.0dev``.

This way it fixes failing build for existing v2.10.x packages.

NOTE: This is pretty much the same change as in https://github.com/StackStorm/bwc-installer/pull/62.

It's "unfortunate" that we repeat that code here. Eventually we should try to refactor it and directly use script from ``bwc-installer`` repo here.


Kami:
Confirmed it's working by re-running two failed builds.

Will go ahead and merge it.

PR #380, opened 2019-04-15
Run BATS tests

blag:
Integrate the RobotFramework/BATS tests rewrite into the usual overall StackStorm end-to-end tests.

Requires StackStorm/st2tests#152 to be merged first.


actions/setup_e2e_tests.sh
```diff
@@ -12,10 +12,24 @@ if [[ -n "$RHTEST" ]]; then
...
+        git clone https://github.com/bats-core/bats-core.git
+        (cd bats-core; sudo ./install.sh /usr/local)
+    fi
 elif [[ -n "$DEBTEST" ]]; then
-    echo "*** Detected Distro is ${DEBTEST} ***"
-    sudo apt-get install -y wget
-    sudo apt-get -q -y install python-pip python-dev build-essential
+    DEBVERSION=`lsb_release --release | awk '{ print $2 }'`
+    echo "*** Detected Distro is ${DEBTEST} - ${DEBVERSION} ***"
+    sudo apt-get -q -y install build-essential jq python-pip python-dev python3-venv wget
```
Kami: Just curious - why do we need python3 here?

actions/setup_e2e_tests.sh
```diff
@@ -12,10 +12,24 @@ if [[ -n "$RHTEST" ]]; then
...
+    if [[ "$DEBVERSION" != "14.04" ]]; then
+        sudo apt-get -q -y install bats
+    else
+        # Install from GitHub
+        git clone https://github.com/bats-core/bats-core.git
```
Kami: Perhaps for consistency, we should just install from git everywhere

actions/st2_chatops_e2e_tests_with_slack.meta.yaml
```diff
@@ -0,0 +1,21 @@
...
+enabled: true
+runner_type: orquesta
+entry_point: workflows/st2_chatops_e2e_tests_with_slack.yaml
+parameters:
+  cwd:
+    type: string
+    default: "/tmp/st2tests"
+  slack_bot_api_token:
+    type: string
+    default: "{{ st2kv.system.chatops_e2e_tests_slack_bot_api_token }}"
```
Kami: While at it - we should probably also utilize encrypted datastore values (should be configured in ops-infra).

Just to dog food that functionality since I don't think we use it anywhere yet in our workflows.

actions/workflows/st2_chatops_e2e_tests.yaml
```diff
@@ -0,0 +1,110 @@
...
+  # This is separated from the configure action because it does not (and SHOULD
+  # not be) run with root privileges
+  setup:
+    action: st2cd.st2_chatops_setup_e2e_tests
+    input:
+      cwd: <% ctx().cwd %>
+    next:
+      - when: <% succeeded() %>
+        do:
+          - run_with_slack
```
Kami: So it looks like we can't run those 3 different adapter tests in parallel?

actions/setup_e2e_tests.sh
blag: We don't - I'll remove it. I originally put it in to test on Ubuntu 14/Cent OS 6 but we're skipping those so I don't need it.

actions/setup_e2e_tests.sh
blag: Once we drop Ubuntu 14 and CentOS 6 we don't need to install from Git anymore.

I'll install from GitHub and leave a comment to that effect. :+1:

actions/st2_chatops_e2e_tests_with_slack.meta.yaml
blag: Agreed.

actions/workflows/st2_chatops_e2e_tests.yaml
blag: The MS Teams and Mattermost tasks are just `core.noop` actions, so I don't think they are going to slow the workflow down any appreciable amount.

Since I haven't written the end-to-end tests for the other adapters, I'm also not sure if they can be parallelized yet, so I figured the placeholders can run in series.

I absolutely will keep parallelization of the distinct end-to-end adapter tests in mind when I am writing them.

blag:
Pushed fixes for the issues @Kami raised, and included a policy to ensure multiple ST2 IUT ("instance under test") do not try to talk with one another during the ChatOps end-to-end tests with Slack.

The concurrency issue will likely contribute to the runtime of these tests by a few minutes (total). I plan on refactoring soon in the future to workaround this issue, but for now the ChatOps end-to-end tests only themselves run for about 2 minutes.

actions/st2_chatops_run_e2e_tests_with_slack.sh
```diff
@@ -0,0 +1,16 @@
...
+#     for RHEL7 systems
+#   - Unnecessary for systems with Python 2.7.9+ (eg: Ubuntu 16.04 and later)
+# * SLACK_USER_USERNAME
+#   - the Slack username for the Python script impersonating a user
+#   - this should be set to the same username as the SLACK_USER Slackbot, below
+# * SLACK_USER_API_TOKEN
+#   - the Slack API token for the Python script that impersonates a user
+#   - THIS MUST BE DIFFERENT THAN SLACK_BOT_API_TOKEN
+
+. venv/bin/activate && python chatops/test_e2e_with_slack.py
```
Kami, line 16: I guess as an improvement we could also use nosetests with timer plugin to provide a bit more detailed output, but this works as well.

Kami (approved):

actions/workflows/st2_chatops_e2e_tests.yaml
```diff
@@ -2,109 +2,140 @@
...
+  # Run the tests with Slack, ST2 IUT on Ubuntu 16.04
+  run_with_slack_u16:
     action: st2cd.st2_chatops_e2e_tests_with_slack
     input:
+      host_ip: <% ctx().host_ip %>
       cwd: <% ctx().cwd %>
-      env:
-        - websocket: <% ctx().websocket %>
+      websocket: <% ctx().websocket %>
+      slack_channel: "{{ st2kv.system.chatops_e2e_tests_slack_channel_u16 }}"
```
Kami, line 87: Let's put a big note somewhere (probably st2tests repo) that Slack bot user needs to be invited to those channel for tests to work.

Kami: And unlike our old tests - I assume those would break and fail loudly if the bot is not invited to those channels and tests fail because of that :)

blag: Yes, these tests are end-to-end, so they will explicitly fail if the bots are not invited to the appropriate channels (and subsequently cannot send or receive messages to/from them).

PR #381, opened 2019-04-17
Add CI for ST2CD

LindsayHill:


LindsayHill:
Trying to work around the rules that rely on githubwebhook.github_event trigger being present. Seems I can't just easily create that trigger here either, as validation will fail for having wrong pack field.

Kami:
We have similar issue with some other packs (chicken and the egg problem with packs which rely on resources from other packs).

One possible work around for that would be to add a special flag ``st2-check-register-pack-resources`` script which would specify a list of resource refs for which we should ignore the errors.

Or perhaps, we could something like this:

1. Register resources first not using  ``--register-fail-on-failure`` flag
2. Register resources again and use ``--register-fail-on-failure`` flag

In theory this would mean that potentially "bad" resources would get registered on the first run, but on the second run it would throw an error if there are any legit issues. I would need to test and confirm that behavior though.

PR #383, opened 2019-04-23
Update st2_prep_release_for_st2 action so it also updates version in bundled packs pack.yaml file

Kami:
It looks like we need another change here - https://github.com/StackStorm/st2cd/pull/369.

Sadly there is a lot of duplicated code all over release flows and scripts (not something I can fix atm...).

### DIFF
--- a/actions/st2_prep_release_for_st2.sh
+++ b/actions/st2_prep_release_for_st2.sh
@@ -106,6 +106,46 @@ do
     fi
 done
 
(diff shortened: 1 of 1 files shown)


Kami:
Will merge it and test it when I re-run the release workflows.

Kami:
Pushed another fix and confirmed it's now working - https://github.com/StackStorm/st2/commit/9a1351418c12de73734f41b51b02007644ee408d

PR #347, opened 2018-10-11
Orquesta migration

blag:
* [x] Run `orquestaconverter` on Mistral workflows
* [x] Manually convert Mistral workflows that could not be autoconverted
* [x] Convert workflows with `with-items` and (possibly) `concurrency`
* [x] Rebase (with git) and manually merge in PRs/changes from `master`
* [x] Add pack icon similar to (but different from) st2ci and st2cicd
* [x] Figure out why the WinRM instance is not coming up
  - [x] Cherry-pick [`f74847d54b123e7892fbbbf5d8f5d82a0f250e63`](https://github.com/StackStorm-Exchange/stackstorm/aws/pull/f74847d54b123e7892fbbbf5d8f5d82a0f250e63) into stackstorm-aws
  - [x] Tag `v0.10.2`
* [x] Rebase everything off of `master`

PR #385, opened 2019-04-26
Add missing prep_st2docs task to st2_prep_dev workflow.

Kami:
Noticed this task was missing while doing the release.

It looks like it was removed (https://github.com/StackStorm/st2cd/pull/347/files#diff-342a6b3541e3f2f4ff8f96cb794b7ce3L66), but never added back during Orquesta migration.


Kami:
Confirmed it's working - https://github.com/StackStorm/st2docs/commit/676885ed4dc747a6a44697b2f3ee4b918910e315.

Merging.

PR #390, opened 2019-05-03
Comment out bats related testing due to failures

warrenvw:
Failure of Bats tests results in build failures. Temporarily commenting them out until failures can be investigated.

```
1..40
ok 1 st2 version works
ok 2 st2 usage works
ok 3 st2 help works
not ok 4 action list for core.local and core.remote action
#
```

and also:

```
/usr/local/libexec/bats-core/bats: line 56: cd: docs: No such file or directory
```

For example, see execution `5ccc3ed8fa307104a3c6852b` and `5ccc3ed8fa307104a3c6852b`.

### DIFF
--- a/actions/workflows/st2_e2e_tests.yaml
+++ b/actions/workflows/st2_e2e_tests.yaml
@@ -114,35 +114,35 @@ tasks:
       - when: <% succeeded() %>
         do:
           - docs_tests
-  docs_tests:
-    action: core.remote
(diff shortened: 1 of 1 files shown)


LindsayHill (approved):

Kami:
@blag It looks like this needs to be re-enabled once all the bats tests are hooked up and confirmed working, right?

PR #393, opened 2019-05-16
Remove old and unused resources

Kami:
This repo is hard to navigate around because it has tons of actions and other resources.

This pull request performs a cleanup and removes some old actions and resources which haven't been relevant for years anymore (AIO installer stuff, puppet stuff, workroom stuff, etc.).


arm4b (approved):
It's always good to see effort on making this repo actual/up-to-date :+1:

Kami:
And in case I removed something we still use, we will know soon enough once we deploy it :P

I think there are a lot more unused actions there, but in the first pass I went with the easy ones - aka ones which haven't been used for years.

PR #395, opened 2019-05-20
Fix change version script files

Kami:
I noticed during the release that the code doesn't work correctly - we need to disable fail on failure functions for steps where we rely on grep exiting with non-zero status code.


Kami:
Confirmed it's working - https://github.com/StackStorm/st2/commit/e5c49f865e2ad15533d973b796d33a0437efa55f.

PR #397, opened 2019-05-21
Decrease ttlMonitorSleepSecs MongoDB setting value for end to end tests

Kami:
This pull request sets ``ttlMonitorSleepSecs`` MongoDB server value to 1 second for end to end tests.

By default MongoDB scans tables and expires items with TTL every 60 seconds. This is not ideal for out end to end tests which test expiry functionality because it means we need to sleep for at least 60 seconds.

This pull request decreases those value so we can speed tests up and make them faster by 60 seconds.

### DIFF
--- a/actions/setup_e2e_tests.sh
+++ b/actions/setup_e2e_tests.sh
@@ -7,7 +7,6 @@ BRANCH=`echo ${VERSION} | cut -d "." -f1-2`
 # Install OS specific pre-reqs (Better moved to puppet at some point.)
 DEBTEST=`lsb_release -a 2> /dev/null | grep Distributor | awk '{print $3}'`
 RHTEST=`cat /etc/redhat-release 2> /dev/null | sed -e "s~\(.*\)release.*~\1~g"`
-SUBTYPE=`lsb_release -a 2>&1 | grep Codename | grep -v "LSB" | awk '{print $2}'`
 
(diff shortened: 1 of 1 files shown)


blag (approved):

actions/setup_e2e_tests.sh
```diff
@@ -8,20 +8,20 @@ BRANCH=`echo ${VERSION} | cut -d "." -f1-2`
 DEBTEST=`lsb_release -a 2> /dev/null | grep Distributor | awk '{print $3}'`
 RHTEST=`cat /etc/redhat-release 2> /dev/null | sed -e "s~\(.*\)release.*~\1~g"`
 
+# Decrease interval for MongoDB TTL expire thread. By default it runs every 60 seconds which
+# means we would need to wait at least 60 seconds in our key expire end to end tests.
+# By decreasing it, we can speed up those tests
+# TODO: Use db.adminCommand, but for that we need to fix admin user permissions in bootstrap script
+echo -e "\nsetParameter:\n  ttlMonitorSleepSecs: 1" >> /etc/mongod.conf
```
Kami, line 15: NOTE: Ideally we could use ``mongo --eval`` and ``db.adminCommand`` here, but it turns out the admin user we create during installer script run doesn't actually give root permissions to admin user so it doesn't have access to run this command...

blag: 🤦‍♂

PR #400, opened 2019-05-23
Change the schedule for the daily CI jobs

m4dcoder:
Change the CI jobs to test stable enterprise packages to 7 AM PST.

### DIFF
--- a/rules/st2_pkg_test_stable_rhel6_enterprise.yaml
+++ b/rules/st2_pkg_test_stable_rhel6_enterprise.yaml
@@ -8,7 +8,7 @@ trigger:
   type: core.st2.CronTimer
   parameters:
     timezone: US/Pacific
-    hour: 9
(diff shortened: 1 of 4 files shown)


arm4b:
@m4dcoder What is the idea behind changing the daily CI jobs schedule to 2 hours earlier?

I'm afraid they can get less attention from someone @oncall since there is a chance for those failure messages to get lost in other Slack conversations.

m4dcoder:
The current schedule, every 3 hours starting from 6 AM and lasting to 4 PM, leaves very little time to do release/maintenance/troubleshooting w/o being in conflict with the scheduled jobs.  I push the intervals down to every 1 hours so the jobs will run from 6 AM and completes by 10 AM.  With a bit of getting used to, I think @oncall should be fine in filtering from the noise.

PR #402, opened 2019-05-24
Increase timeout of test_quickstart_key action

### LINKED ISSUE #401: test_quickstart_key action times out on RHEL6 and RHEL7
On RHEL6 and RHEL7 distros, the test_quickstart_key action which gets run as part of st2cd.st2_e2e_tests_test_quickstart - times out. Due to this main run_e2e_tests actions fails.

VineeshJain:
Closes #401 

This PR Increases the timeout of test_quickstart_key action to 90s. Currently the default timeout of 60s is used which is causing the tests to timeout on RHEL 6 and RHEL 7, though the actions are successfully completed on the remote VM.


m4dcoder (approved):
LGTM

PR #407, opened 2019-07-02
Fix dev version assignment in docker_prep_dev

m4dcoder:
Fix docker_prep_dev so that it will take the given next release x.y.z (i.e. 3.2.0) and assign the dev version to be x.ydev (i.e. 3.2dev).


arm4b (commented):
:+1:

PR #408, opened 2019-07-15
Add rules to run CI for stable u18 daily

m4dcoder:
Add timer rules to run daily CIs for stable community and enterprise u18 packages. Add u18 as supported distro when updating these rules during release.


arm4b (approved):
Thanks for remembering and fixing this after release :+1:

m4dcoder:
Of course. Whether I have time to do all this is a different thing...

PR #418, opened 2019-08-19
Use unencrypted st2kv tokens for e2e tests

arm4b:
Following https://github.com/StackStorm/ops-infra/pull/382


blag:
NAK - these should be stored encrypted, which means they need to be decrypted, right?

This workflow worked perfectly on the previous incarnation of st2cicd, so (AIUI) this is not the right place to fix the issue. Let's make sure the encrypted values were stored in the datastore and decrypt them here.

blag (changes requested):
NAK

blag (approved):
Approved pending changes to our Ansible code.

Thanks!

PR #424, opened 2020-03-05
Disable RHEL 8 nightlies until we actually have a stable release for RHEL 8

blag:
Because this currently just causes unnecessary nightly build failures.


m4dcoder (approved):
LGTM

punkrokk:
Makes sense. Sorry it’s causing so many failures.

punkrokk (approved):

PR #427, opened 2020-03-14
Use st2tests branch

blag:
The CI/CD process has previously only been able to clone from two branches of the StackStorm/st2tests repository: `master` and a version tag.

That makes it really difficult to test changes to st2tests, like StackStorm/st2tests#185. That means that I would have to merge that into `master` and rerun the CI/CD workflows to test with the `master` branch of st2tests. And if things broke, I would have to do the exact same thing again.

Instead of doing that, we should explicitly expose the branch of st2tests that will be cloned. This allows us to manually test out our st2tests changes against the current `master` branch of ST2 _before_ we merge those changes into st2tests `master`.

This should hopefully drive better software practices, as one more piece of the CI/CD system can be tested in isolation to the other pieces.

A corresponding PR is incoming for st2ci.

And one more thing: this change also enables us to add CI to the st2tests repository in the future, by adding a rule to run this with the st2tests PR branch and configuring that in GitHub.

### DIFF
--- a/actions/setup_e2e_tests.meta.yaml
+++ b/actions/setup_e2e_tests.meta.yaml
@@ -8,7 +8,9 @@ parameters:
   version:
     type: string
     default: "master"
-    description: Version used for the release. Should include the patch e.g. 0.1.0
+    description: >
+      Branch name or version string used for the release. If a version string,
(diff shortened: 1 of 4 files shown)


nmaludy (approved):
LGTM!

PR #429, opened 2020-04-09
Fix YAQL comparison operators for distro check

blag:
Fixes operators when checking Linux distribution for Mistral support. I messed this up when I added RHEL 8 support to this workflow.

Verified this works by installing this branch on st2cicd and manually rerunning a previously failed `st2ci.st2_pkg_e2e_test` workflow with `distro: RHEL8`.


nmaludy (approved):
LGTM. Could probably reduce computations by assigning `ctx().host_fqdn.toLower()` to its own var.

punkrokk (approved):
LGTM

PR #430, opened 2020-04-10
Rely on a newer 'slack.chat.postMessage' instead of using old 'slack.post_message'

arm4b:
This will make st2cd repo consistent with the https://github.com/stackstorm/st2ci about using the same Slack action.


blag (commented):
:+1: Thanks!

PR #432, opened 2020-04-16
Fix run_e2e_tests to skip Mistral and properly not-skip everything else

blag:
This bug is pretty straightforward. One `when` condition is incorrectly checking for `rhel8` in the host FQDN, while the other `when` condition is correctly checking for `el8` in the host FQDN.


m4dcoder (approved):
LGTM

blag:
Here is a live action replay of me looking at the blame information for these files:

spiderman-pointing-at-self

punkrokk:
Can we memorialize this somewhere as an Easter egg?

arm4b:
:+1:

PR #439, opened 2020-07-10
Deprecate RHEL 6 / CentOS 6

amanda11:
Remove EL6 builds

addressing st2cd - remove EL6 from StackStorm/discussions#39


amanda11:
A few files where I added EL8 references, as weren't sure why they were missing - but need careful review:

- CHECKLIST.md
- actions/st2_prep_release_cd_rules.meta.yaml
- actions/st2_prep_release_rules.meta.yaml
- actions/workflows/flow_build.yaml

In addition, there is still a file: ./rules/st2cd_slack_pkg_rhel6.yaml - I wasn't quite sure where this rule fit in, as we didn't have one for el7 or el8 - just seem to have one slack rule. So wasn't sure if this needed amending.

actions/setup_e2e_tests.sh
```diff
@@ -59,11 +56,6 @@ if [[ -n "$RHTEST" ]]; then
         rm -rf bats-core
     fi
```
blag, line 54: Just removing these lines will break pretty much all of the ChatOps end-to-end tests, since they rely on the `bats` command being available. Instead of installing `bats` from its GitHub repo, we should install BATS from its RPM package with YUM. You might want to double check the repos for the exact name, but I believe it's just `bats`.

blag (changes requested):
Looking great (except for that one place)!

amanda11:
@blag Is there any follow-up required on this, I think I've addressed your review comment.

blag:
Checking on one last thing before merging.

actions/setup_e2e_tests.sh
```diff
@@ -41,29 +41,19 @@ if [[ -n "$RHTEST" ]]; then
...
+        sudo yum install -y python3-pip wget jq 
         PIP="pip3"
-    fi
-
-    # Remove bats-core if it already exists (this happens when test workflows
-    # are re-run on a server when tests are debugged)
-    if [[ -d bats-core ]]; then
+        # bats not available in epel for EL 8, Install from GitHub
         rm -rf bats-core
+        git clone --branch add_per_test_timing_information --depth 1 https://github
```
blag: This URL is not complete.

actions/setup_e2e_tests.sh
```diff
@@ -41,29 +41,19 @@ if [[ -n "$RHTEST" ]]; then
...
         # For RHEL/CentOS 8 and above
-        sudo yum install -y python3-pip wget jq
+        sudo yum install -y python3-pip wget jq 
         PIP="pip3"
-    fi
-
-    # Remove bats-core if it already exists (this happens when test workflows
-    # are re-run on a server when tests are debugged)
-    if [[ -d bats-core ]]; then
+        # bats not available in epel for EL 8, Install from GitHub
```
blag: This is going to sound weird. You're right that bats is not available in EPEL for RHEL 8. But it is available, for some reason, via [npm](https://github.com/bats-core/bats-core#npm). So instead of installing from GitHub, let's install with npm:

```bash
npm install --global bats
```

blag (changes requested):

actions/setup_e2e_tests.sh
amanda11: Odd - will fix

amanda11:
@blag Should have addressed your review comments, thanks for spotting.

blag (approved):

blag:
Thanks!

PR #440, opened 2020-08-03
Remove mistral from st2cd

amanda11:
Remove mistral from st2cd

Addresses st2cd entry in https://github.com/StackStorm/st2/issues/4762


amanda11:
I couldn't test installation of this pack as couldn't find the pack githubwebhook which was referenced.

Also there is an UPDATE_MISTRAL parameter passed to st2_chg_ver_for_st2.sh. I've left the parameter in, but it has no affect.
 I can remove this and change the order of parameters, but I wasn't sure where this workflow was called from to know if that was safe to do.

blag (approved):

nmaludy (approved):

punkrokk (approved):
LGTM

PR #442, opened 2020-10-19
Enhanced error message when st2_prep_release is run twice

nmaludy:
When working on the `3.3` release i accidentally ran the `st2_prep_release` action too soon (going to fix this by editing the Release process Wiki). Coming back to this a week or so later, i thought about running this action again and came across a bad error message that told me what was wrong, but didn't tell me what i should be doing instead.

I've edited the error message with some helpful guidance to the next person performing a release, so hopefully they won't stub their toe like i did.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,13 @@
+# Changelog
+
+## 0.1.3
+
+Added descriptive error message to `st2cd.st2_prep_release_for_st2` in case that workflow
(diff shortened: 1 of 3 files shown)


arm4b (approved):
Makes sense and helps future generations of Stormers hitting the same issue :+1:

blag (approved):
That's a looooong single line, but it's a great error message now!

PR #445, opened 2020-10-28
Fix StackStorm capitalization

blag:
During the 3.3.0 release, we ran into issues where the CircleCI API didn't use to be case-sensitive but it is now. As such, we need to properly capitalize "StackStorm" when we specify the `username` parameter to `circle_ci.run_build` tasks.

Additionally, this issue also affects the `st2cd.sync_api_docs_on_st2_commits` rule during normal development.

This PR implements that change in the rule and the workflows to stage and release packages, and pulls the username into a workflow variable so we don't need to repeat ourselves.

Bonus: I also spaced out the workflow root keys a bit to improve the readability.

### DIFF
--- a/actions/workflows/bwc_stage_release_packages.yaml
+++ b/actions/workflows/bwc_stage_release_packages.yaml
@@ -13,7 +13,7 @@ input:
       UBUNTU18: u18
 
 vars:
-  - username: stackstorm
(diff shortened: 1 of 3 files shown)


nmaludy (approved):
LGTM

PR #446, opened 2020-12-02
Use Python 3 to create virtualenvs

blag:
This tweaks `setup_e2e_tests.sh` to install Python 3 pip and virtualenv packages for both RHEL 7 & 8 and Ubuntu tests.


blag:
~Waiting on StackStorm/st2-packages#679.~

This is independent of StackStorm/st2-packages#679, although tests may still fail until that is merged.

cognifloyd:
Oh. I guess this PR obsoletes #452

PR #449, opened 2020-12-12
Add st2tests to prep_release

anirudhbagri:
Add st2tests to prep_release


arm4b (changes requested):
Thanks a lot! :+1: 

Quick search identified another place required for update:
https://github.com/StackStorm/st2cd/blob/master/actions/st2_chg_ver_for_st2.sh#L49-L59
Can you please also include changes for that one?

anirudhbagri:
Added.

arm4b (approved):
:+1:

PR #450, opened 2020-12-13
Use Orquesta task retry

blag:
This PR updates our workflows to use Orquesta task retry instead of implementing that ourselves.


arm4b (approved):
:+1:

blag:
Waiting on #194 and StackStorm/st2#5090 - I want to get Python 3 stuff merged ASAP before I go around potentially breaking the build process with this.

PR #451, opened 2021-03-03
Update release workflows

blag:
Fix workflows and scripts for the v3.4 release, which required changes to the way the LDAP authentication and RBAC authorization backend plugins are handled.

This includes removing some now-useless bwc workflows, and making the `setup_e2e_tests.sh` script idempotent, which is very important when you are trying to debug a ST2 instance.

The ChatOps end-to-end tests were skipped on Ubuntu Xenial since they were running into a very weird recursive logging issue - only on Ubuntu Xenial - that we did not see any trace of in user testing. Ubuntu Xenial is EOL in ~1 month anyway, so in the unlikely event that the bug is valid, it is still unlikely that any users will hit it, and if they do, it would be silly for us to support ST2 on an OS that has reached EOL status anyway.

Long live Python 3.6+!


blag:
~This is also a WIP - please DO NOT MERGE.~

actions/setup_e2e_tests.sh
```diff
@@ -122,21 +131,51 @@ sudo cp -R st2tests/packs/* /opt/stackstorm/packs/
...
+    PIP_VERSION=$(curl --silent https://raw.githubusercontent.com/StackStorm/st2/${BRANCH}/Makefile | grep 'PIP_VERSION ?= ' | awk '{ print $3 }')
+else
+    PIP_VERSION=$(curl --silent https://raw.githubusercontent.com/StackStorm/st2/v${VERSION}/Makefile | grep 'PIP_VERSION ?= ' | awk '{ print $3 }')
+fi
+# Fallback
+if [[ -z "$PIP_VERSION" ]]; then
+    PIP_VERSION=$(curl --silent https://raw.githubusercontent.com/StackStorm/st2/master/Makefile | grep 'PIP_VERSION ?= ' | awk '{ print $3 }')
+fi
+sudo ${PIP} install --upgrade "pip==$PIP_VERSION"
 sudo ${PIP} install --upgrade "virtualenv==15.1.0"
```
amanda11: Should we be getting version of virtualenv from the fixed-requirements.txt in st2? I'm not sure why this is so low... It was 16.6.0 in the st2 fixed-requirements until recently..

actions/setup_e2e_tests.sh
```diff
@@ -122,21 +131,51 @@ sudo cp -R st2tests/packs/* /opt/stackstorm/packs/
...
+# other operating systems), we simply extend PATH with $HOME/.local/bin to help
+# Bash find the virtualenv executable.
+# Also, it's 2021, but we still have to tell virtualenv to configure the
+# virtualenv to use Python 3.
+PATH=$PATH:$HOME/.local/bin virtualenv --no-download --python=$PYTHON3 venv
 . venv/bin/activate
+# Set pip and virtualenv within the virtualenv to ensure the Python 3-only
+# dependencies can be successfully installed
+${PIP} install --upgrade "pip==$PIP_VERSION"
+${PIP} install --upgrade "virtualenv==15.1.0"
```
amanda11: Same comment as above about virtual env version

actions/workflows/st2_chatops_e2e_tests.yaml
```diff
@@ -105,16 +105,23 @@ tasks:
...
+    # 5. Ubuntu Xenial is EOL in ~1 month anyway
+    action: core.noop
+    # action: st2cd.st2_chatops_e2e_tests_with_slack
+    # input:
+    #   host_ip: <% ctx().host_ip %>
+    #   st2_auth_token: <% ctx().env.ST2_AUTH_TOKEN %>
+    #   websocket: <% ctx().websocket %>
+    #   slack_channel: <% ctx().chatops_e2e_tests_slack_channel_u16 %>
+    #   slack_bot_username: "{{ ctx().chatops_e2e_tests_slack_bot_username_u16 }}"
+    #   slack_bot_api_token: "{{ ctx().chatops_e2e_tests_slack_bot_api_token_u16 }}"
```
amanda11: I thought following the recent change we could put this test back in?

amanda11 (commented):
Looks great - I had a few comments that you may or may not agree with.

actions/workflows/st2_chatops_e2e_tests.yaml
blag: Good call, I'll do that.

actions/setup_e2e_tests.sh
blag: I forgot to respond to this, sorry!

I'm looking at this code again, and I'm not sure where we even use the created virtualenv.

I'm happy to update this to pull the virtualenv version of st2's `fixed-requirements.txt` or similar file, but at this point I don't want to expand this PR any further.

However, before we look into that, we should double check that this virtualenv is actually used somewhere.

Wayyy back in the day we used RobotFramework to test ChatOps, but I ripped those out and replaced them with BATS tests, see:

* https://github.com/StackStorm/st2cd/blob/master/actions/workflows/st2_chatops_e2e_tests.yaml
* https://github.com/StackStorm/st2cd/blob/master/actions/workflows/st2_chatops_e2e_tests_with_slack.yaml
* https://github.com/StackStorm/st2cd/blob/master/actions/workflows/st2_e2e_tests.yaml (at the bottom)

And our current code runs the ChatOps tests [using the virtualenv for the `chatops_tests` pack](https://github.com/StackStorm/st2cd/blob/master/actions/workflows/st2_chatops_e2e_tests_with_slack.yaml#L87):

```yaml
  # Run the tests with Slack
  # Default SLACK_WAIT_FOR_MESSAGES_TIMEOUT value is 120 (seconds)
  run_with_small_timeout:
    action: core.remote
    input:
      hosts: <% ctx().host_ip %>
      timeout: 2400  # seconds
      # Workaround for https://github.com/StackStorm/st2/issues/4676
      # cmd: st2 run --inherit-env tests.test_aliases_with_slack
      # env:
      #   ST2_AUTH_TOKEN: <% ctx().st2_auth_token %>
      #   WEBSOCKET_CLIENT_CA_BUNDLE: <% ctx().websocket %>
      #   SLACK_CHANNEL: <% ctx().slack_channel %>
      #   SLACK_BOT_USERNAME: <% ctx().slack_bot_username %>
      #   SLACK_USER_USERNAME: <% ctx().slack_user_username %>
      #   SLACK_USER_API_TOKEN: <% ctx().slack_user_api_token %>
      cmd: |
        export ST2_AUTH_TOKEN=<% ctx().st2_auth_token %>;
        export WEBSOCKET_CLIENT_CA_BUNDLE=<% ctx().websocket %>;
        export SLACK_CHANNEL=<% ctx().slack_channel %>;
        export SLACK_BOT_USERNAME=<% ctx().slack_bot_username %>;
        export SLACK_USER_USERNAME=<% ctx().slack_user_username %>;
        export SLACK_USER_API_TOKEN=<% ctx().slack_user_api_token %>;
        export SLACK_WAIT_FOR_MESSAGES_TIMEOUT=240;
        . /opt/stackstorm/virtualenvs/chatops_tests/bin/activate;
        python /opt/stackstorm/packs/chatops_tests/actions/test_aliases_with_slack.py
```

So I think this entire pip+virtualenv is now superfluous and can be removed.

However, this PR is big enough at this point - I'd like to get it merged in more than I'd like it to solve All Of The Things. ¯\\\_(ツ)\_/¯

amanda11: Fair enough - something to look at separate to this PR.

blag: Captured in #454.

blag: I implemented this, but we should still look at whether or not it's useful or necessary.

actions/workflows/st2_chatops_e2e_tests.yaml
blag: Done. :+1:

PR #452, opened 2021-03-06
Use the same version of pip as st2 core

cognifloyd:
st2 core uses pip 20.0.2. Do the same here.


Kami (approved):

blag:
On mobile right now so I can't easily link to it, but I think I covered this better in #451, where I pull the version of out st2's Makefile.

Same pattern that I think we do in st2packaging-dockerfiles or someplace else.

cognifloyd:
Yeah. This is better resolved by #451 and/or #446.

PR #453, opened 2021-03-30
Need to enable TLS else the downloadfile fails complaining about TLS/SSL

amanda11:


blag (approved):

blag:
I don't know Powershell, but this looks reasonable to me. You might want to get another signoff though.

Issue #438, opened 2020-05-29
Update Release Process for Open Sourced EWC


arm4b:
Due to historical reasons, the Release Automation Workflows in this repo include both StackStorm OSS and EWC Enterprise.

As Extreme Networks open-sourced the EWC Enterprise (https://stackstorm.com/2020/05/27/extreme-networks-donates-ewc-to-linux-foundation/), we'll need to cleanup and rework all the automation and workflows to make them actual with the new environment.

While the description looks clear enough, this may be one of the rabbit holes in the [EWC OpenSource](https://github.com/orgs/StackStorm/projects/14) plan.

arm4b:
Per @m4dcoder call, there are Workflows that need to be modified to test the LDAP/RBAC functionality:
https://github.com/StackStorm/st2cd/blob/master/actions/workflows/st2_enterprise_tests.yaml and others.

Additionally, in `st2` repository https://github.com/StackStorm/st2/blob/e5131b8921a5fac27307c0cd292bcfa99a56ba31/requirements.txt#L20-L21 needs to be modified to use the branch like `3.3`, `3.2`, instead of `master` for each release.

The automation for branch creation is already in place:
- https://github.com/stackstorm/st2-rbac-backend/tree/v3.2
- https://github.com/StackStorm/st2-auth-ldap/tree/v3.2

blag:
Status update: this was partially completed in #451.

## Done ##

The `st2_enterprise_tests` action was removed in #451.

* [x] The [`verify_stanley_ssh_key_is_generated_and_installed`](https://github.com/StackStorm/st2cd/pull/451/files#diff-f973f498b17b234eaf09b73e23ccdae4defa736a7bb750a062cf60fa69eeb940L6) test is obsolete since previous tests will fail if the `stanley` key is not installed.
* [x] The [`verify_rbac_is_enabled`](https://github.com/StackStorm/st2cd/pull/451/files#diff-f973f498b17b234eaf09b73e23ccdae4defa736a7bb750a062cf60fa69eeb940L16) test is partially obsolete since it checks for `rbac.backend = enterprise` which doesn't work anymore.
[...]

PR #456, opened 2021-05-04
Update prep release to update st2 requirements files for ldap and rbac

amanda11:
Update the prep release so that it also updates the requirement files to point to the relevant branch for st2 auth ldap and st2 rbac backend.
Part of https://github.com/StackStorm/st2cd/issues/438

### DIFF
--- a/actions/st2_prep_release_for_st2.sh
+++ b/actions/st2_prep_release_for_st2.sh
@@ -116,9 +116,9 @@ done
 BUNDLED_PACKS_METADATA_FILES=($(find contrib/ -mindepth 2 -maxdepth 2 -name pack.yaml))
 
 # Set version attribute for all the requirements.txt that refer to st2-auth-ldap
-LDAP_REQUIREMENT_FILES=($(find ./ -maxdepth 2 -name "*requirements.txt" | xargs grep -l "egg=st2-auth-ldap"))
-# Set version attribute for all the requirements.txt that refer to st2-auth-ldap
-RBAC_REQUIREMENT_FILES=($(find ./ -maxdepth 2 -name "*requirements.txt" | xargs grep -l "egg=st2-rbac-backend"))
+LDAP_REQUIREMENT_FILES=($(find ./ -maxdepth 2 -name "*requirements.txt" | xargs egrep -l '\bst2\b.*\bldap\b'))
+# Set version attribute for all the requirements.txt that refer to st2-rbac-backend
+RBAC_REQUIREMENT_FILES=($(find ./ -maxdepth 2 -name "*requirements.txt" | xargs egrep -l '\bst2\b.*\brbac\b'))
 
 # Temporary disable fail on failure for grep step where failure is OK
 set +e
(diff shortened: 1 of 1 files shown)


actions/st2_prep_release_for_st2.sh
```diff
@@ -115,6 +115,11 @@ done
 # Set version attribute for all the bundled packs (core, linux, examples, etc.)
 BUNDLED_PACKS_METADATA_FILES=($(find contrib/ -mindepth 2 -maxdepth 2 -name pack.yaml))
 
+# Set version attribute for all the requirements.txt that refer to st2-auth-ldap
+LDAP_REQUIREMENT_FILES=($(find ./ -maxdepth 2 -name "*requirements.txt" | xargs grep -l "egg=st2-auth-ldap"))
+# Set version attribute for all the requirements.txt that refer to st2-auth-ldap
+RBAC_REQUIREMENT_FILES=($(find ./ -maxdepth 2 -name "*requirements.txt" | xargs grep -l "egg=st2-rbac-backend"))
```
blag: There's no guarantee that these will continue to be distributed as eggs AFAIK. So I would adjust the `grep` to target a wider range of values:

```suggestion
LDAP_REQUIREMENT_FILES=($(find ./ -maxdepth 2 -name "*requirements.txt" | xargs egrep -l '\bst2\b.*\brbac\b'))
# Set version attribute for all the requirements.txt that refer to st2-auth-ldap
RBAC_REQUIREMENT_FILES=($(find ./ -maxdepth 2 -name "*requirements.txt" | xargs egrep -l '\bst2\b.*\bldap\b'))
```

I'm not 100% sure I got the backslash escaping correctly, so you might want to check this.

It's unlikely that any Python package will contain `st2` and `rbac` or `ldap`.

Also, there's a copy/paste error with the comment:

```suggestion
LDAP_REQUIREMENT_FILES=($(find ./ -maxdepth 2 -name "*requirements.txt" | xargs grep -l "egg=st2-auth-ldap"))
# Set version attribute for all the requirements.txt that refer to st2-rbac-backend
RBAC_REQUIREMENT_FILES=($(find ./ -maxdepth 2 -name "*requirements.txt" | xargs grep -l "egg=st2-rbac-backend"))
```

actions/st2_prep_release_for_st2.sh
```diff
@@ -151,6 +156,48 @@ if [ "${IS_DEV_VERSION}" = "false" ]; then
...
     done
+    for REQUIREMENT_FILE in "${LDAP_REQUIREMENT_FILES[@]}"
+    do
+        echo "Setting ldap branch version in: ${REQUIREMENT_FILE}"
+
+        if [[ ! -e "${REQUIREMENT_FILE}" ]]; then
+            >&2 echo "ERROR: Requirement file ${REQUIREMENT_FILE} does not exist."
+            exit 1
+        fi
+
```
blag: The previous `find` command should have returned all of the files, and those files should exist. Is there any reason that the files would be visible/readable by the `find` command but not here?

```suggestion
```

actions/st2_prep_release_for_st2.sh
```diff
@@ -151,6 +156,48 @@ if [ "${IS_DEV_VERSION}" = "false" ]; then
...
+            echo "Setting ldap version branch in ${REQUIREMENT_FILE} to ${BRANCH}..."
+            sed -i -E "s/^git\+https:\/\/github.com\/StackStorm\/st2-auth-ldap\.git@(.*?)st2-auth-ldap$/git\+https:\/\/github.com\/StackStorm\/st2-auth-ldap\.git@${BRANCH}#egg=st2-auth-ldap/" ${REQUIREMENT_FILE}
+
+            VERSION_STR_MATCH=`grep "st2-auth-ldap.git@${BRANCH}" ${REQUIREMENT_FILE} || true`
+            if [[ -z "${VERSION_STR_MATCH}" ]]; then
+                >&2 echo "ERROR: Unable to update the ldap version in >${REQUIREMENT_FILE}."
+                exit 1
+            fi
+        fi
+    done
```
blag: NAK. Multiple issues with this.

1. Use the `FORK` variable instead of hardcoding `StackStorm` so this respects the existing parameters.
2. Don't match on `#egg=`, see above.
3. Instead of checking for the `BRANCH` before updating it, I would remove the `if` check and "blindly" do the search-and-replace with `sed`. If that ends up not changing anything, then the script will detect that farther down and won't try to commit or push. This keeps the script more simple, easier to understand and easier to edit if/when we switch to releasing `st2-auth-ldap` and `st2-rbac-backend` as their own Python packages and change these lines.
4. Not sure what value the post-`sed` check has.
5. The `>` seems weird in an error message, and makes copying and pasting that filename more difficult. I would remove it, but that's me.

Same story for the RBAC check down below as well.

blag (changes requested):
Coming along nicely, just a few things to reconsider.

amanda11:
@blag  Thanks for review, your find greps were perfect. Review comments hopefully addressed.

blag (approved):

blag:
Looks much simpler than before, thanks for refactoring! 😃

PR #460, opened 2021-05-25
Move test on focal

amanda11:


amanda11:
Working on CI server. Merging.

PR #466, opened 2021-07-13
Change so only build stable builds on Tuesday

amanda11:


arm4b (commented):
I'm afraid it would be hard to monitor the Sunday build.
Monday is the hard day :) How about mid-week CI run like Tuesday?

arm4b (approved):
:+1:

PR #467, opened 2021-10-08
Fix logic that determines whether to add new key or amend existing key

amanda11:
With the introduction of https://github.com/StackStorm/st2-packages/pull/706 the systemd generators will fail to create the st2api.socket etc files if there are duplicate keys in st2.conf.

When setup_e2e_tests.sh is run it was adding duplicating encryption_key_path keys, as it had its if /elif the wrong way round - so if the key was there with different value it woudl always add a new value instead of amending the old one.

This causes the E2E tests to break.


nzlosh (approved):
LGTM

PR #468, opened 2021-11-25
Fix several actions to use double quotes

wingiti:
not sure if they are still all needed. this is related to issue: https://github.com/StackStorm/st2/issues/5455


arm4b (approved):
Good catch and thanks for the fix! :+1:

